PR #8, opened 2023-04-12
Add HTTP enabled bitswap impl

MarcoPolo:
Different than #6, in this once we use bitswap and race against http endpoints. Those http endpoints may be other libp2p peers we've communicated or HTTP gateways (In my mind these are really static buckets/websitesl somewhere).

Using vendored deps for ease of iterating and so other folks can run this as well.


SgtPooki:
@aschmahmann same here? I think we want to stick with trustless gateway and any of this bitswap/http stuff should go in verified-fetch directly?

PR #17, opened 2024-01-24
feat: use @helia/verified-fetch

### LINKED ISSUE #4: bug: fix css font loading issues for libp2p.io
check out https://helia-service-worker-gateway.on.fleek.co/helia-sw/ipns/libp2p.io (make sure the sw is loaded and updated from the root first)

You'll notice a failed request for `Qmf9eUtc1TML1MS3sZxh3DWw9LfJPjuL3h7XSnhR2nGZuW/css/common.css/fonts` in the console, it will say something like:

```
fs.stat error for cid 'Qmf9eUtc1TML1MS3sZxh3DWw9LfJPjuL3h7XSnhR2nGZuW' and path '/css/common.css/fonts'
```

And also 404s in the network tab (cause heliaFetch is throwing):

image

Checking this in the terminal shows the following:

```
╰─ ✔ ❯ ipfs ls Qmf9eUtc1TML1MS3sZxh3DWw9LfJPjuL3h7XSnhR2nGZuW/css/common.css/fonts
Error: func called on wrong kind: "LookupBySegment" called on a data.Bytes node (kind: bytes), but only makes sense on map or list

╰─ ✘ 1 ❯ ipfs ls Qmf9eUtc1TML1MS3sZxh3DWw9LfJPjuL3h7XSnhR2nGZuW/css/common.css


╰─ ✔ ❯ ipfs ls Qmf9eUtc1TML1MS3sZxh3DWw9LfJPjuL3h7XSnhR2nGZuW/css/
QmctPuZeC91dNjDtga2GhVV7wZsddtZo5z5KSVpKM2oMNE 52394 common.css
Qmb8ByzaP9yUBzxYSc5PSJDww8BGj6xKGW6CefyQcmDLHK 14206 implementations.css
QmcXAmwqBY8muaKkUKG5Nd1NUuCrb9ygX6YHU9fi95nxTL 13531 index.css
QmPykMivu7CaR5h7NjwGKsavFGivVapBpNRyyDuSH3fwLA 15258 media.css
```

I'm not sure why we're receiving network requests for this, it could be some dynamic JS stuff for loading of these fonts. If it's a pattern we can recognize consistently (due to some fontawesome convention that servers implement), we may be able to redirect it appropriately, but we may need to simply ignore this.

### LINKED ISSUE #7: feat: export heliaFetch as a separate lib
We should allow users to import heliaFetch directly for use in their service-workers or client-side applications

```[tasklist]
### Tasks
- [ ] Agree on interface for helia-fetch (is helia, path, and fetchOptions enough?)
- [ ] Agree on repo location for helia-fetch (does it make sense to live in the ipfs/helia monorepo?)
```

aschmahmann:
Closes #7
Closes #4


SgtPooki:
Demo video of loading subdomains available at 

CIDV0 = ipfs://QmRhzWfzRHHmMirAQrBimyuiaCRFbPprWLuyw3cBqmHUQy
CIDv1 = ipfs://bafybeibsbktccr2krbjyvoclsynzxghaclnvoiwufxafy5gq6hyuwj5tpi

SgtPooki:
webui via helia service worker from this branch deployed to sw.sgtpooki.com (modified BASE_URL manually): 

https://webui-ipfs-io.ipns.sw.sgtpooki.com/

https://bafybeidf7cpkwsjkq6xs3r6fbbxghbugilx3jtezbza7gua3k5wjixpmba.ipfs.sw.sgtpooki.com

src/index.tsx
```diff
@@ -13,26 +17,30 @@ const container = document.getElementById('root')
...
-const heliaSwPath = urlSearchParams.get('helia-sw')
-if (heliaSwPath != null) {
+const subdomain = window.location.hostname.replace(`.${BASE_URL}`, '')
+const subdomainRegex = /^(?<origin>[^/]+)\.(?<protocol>ip[fn]s)?$/
+const subdomainMatch = subdomain.match(subdomainRegex)
+
+const isPathBasedRequest = window.location.pathname.startsWith('/ip')
+const isSubdomainRequest = subdomainMatch?.groups != null
+
+if ((isPathBasedRequest || isSubdomainRequest)) {
```
2color: How does this check for whether the SW has registered/triggered work?  Wouldn't `isPathBasedRequest || isSubdomainReques` be the same?

src/get-helia.ts
```diff
@@ -1,57 +1,24 @@
...
-  // create a Helia node
-  const helia = await createHelia({
-    datastore: datastore as unknown as HeliaInit['datastore'],
-    blockstore,
-    libp2p
+export async function getHelia (): Promise<Helia> {
+  const config = await getConfig()
+
+  const helia = await createHeliaHTTP({
+    blockstore: new LevelBlockstore('./blockstore'),
```
2color: What's the reason for using `LevelBlockstore` over `blockstore-idb`?

src/lib/config-db.ts
```diff
@@ -0,0 +1,66 @@
+import { Level } from 'level'
```
2color: Why is level needed here? 

Wouldn't it be possible to simplify this by using IndexedDB directly?

src/get-helia.ts
```diff
@@ -1,57 +1,24 @@
...
+    datastore: new LevelDatastore('./datastore'),
+    blockBrokers: [
+      trustlessGateway({
+        gateways: ['https://cloudflare-ipfs.com', 'https://dweb.link', 'https://trustless-gateway.dev', ...config.gateways]
```
lidel: Fix typo + avoid subdomain redirect and CORS preflight for every subdomain block request + default to dedicate trustless domain:
```suggestion
        gateways: ['https://cloudflare-ipfs.com', 'https://trustless-gateway.link', 'https://ipfs.io', ...config.gateways]
```

src/lib/heliaFetch.ts
```diff
@@ -1,17 +1,108 @@
...
+    // no other way to determine file-type.
+    return defaultMimeType
+  }
+
+  // no need to include file-types listed at https://github.com/SgtPooki/file-type#supported-file-types
+  switch (fileName.split('.').pop()) {
+    case 'css':
+      return 'text/css'
+    case 'html':
+      return 'text/html; charset=utf-8'
```
lidel: Q: why are we assuming every HTML uses UTF-8?

src/get-helia.ts
2color: Is there any benefit to having **both** `ipfs.io` and `trustless-gateway.link` given that they're backed by the same infrastructure? 

```suggestion
        gateways: ['https://cloudflare-ipfs.com', 'https://trustless-gateway.link', ...config.gateways]
```

lidel: Yeah, we should only have `https://trustless-gateway.link`

ps. in Google Chrome, the default limit is 6 connections per domain, so having more domains, even when backed by the same backend, may be useful, but only if we utilize them in paralllel and see us hitting the limit often. but if that is the case, we should have `https://1.trustless-gateway.link` etc rather than `ipfs.io` which a burned domain, often blocked or triggering security software.

src/get-helia.ts
SgtPooki: No reason, we could use any from https://github.com/ipfs/js-stores but level-* are the ones that I have seen and end up using. It seems like the recommendation is to use indexDb though. 

> N.b. this is here largely for the sake of completeness, in node you should probably use FSDatastore, in browsers you should probably use IDBDatastore.
> \- https://www.npmjs.com/package/blockstore-level

I'll update this.

src/get-helia.ts
SgtPooki: https://cloudflare-ipfs.com/ always fails due to CORS. I'm going to remove that one and just stick with trustless-gateway.link for now

src/lib/heliaFetch.ts
SgtPooki: bad habit. removed.

src/lib/config-db.ts
SgtPooki: fixed

src/index.tsx
SgtPooki: I added some tests & comments which show more about what is being checked.

If the service worker is registered, and it's a valid path, the request is intercepted by the service worker.

If the service worker is not registered, it will reload the page and the service worker will intercept things after the reload.

There could still be some more intelligent determinism for the sw registration & redirect logic.

SgtPooki:
merging this so we can move forward with a much smaller PR

PR #27, opened 2024-02-22
feat: BASE_URL can be overridden

SgtPooki:


SgtPooki:
verified this works on sw.sgtpooki.com with `BASE_URL="sw.sgtpooki.com" npm run build && scp dist/* root@sgtpooki.com:/var/www/sw.sgtpooki.com/html`

PR #29, opened 2024-02-23
fix: remove invalid char from the helia fetch url

2color:


SgtPooki (approved):

SgtPooki:
nice. I fixed this in my config page PR and didn't see this until I saw your status update.

PR #34, opened 2024-02-24
fix: dnsLinkLabel decode encode

SgtPooki:
fixes an issue with dns link encoding and decoding and adds tests


SgtPooki (commented):
self review.

PR #53, opened 2024-02-26
fix: dynamic subdomain gateway detection

lidel:
## Description

This PR removes the concept of hardcoding `BASE_URL` at build time, which makes the same static payload work on any domain that follows subdomain convention, including inlined DNSlink names required for TLS wildcard setups:

- http://docs.ipfs.tech.ipns.helia-sw-gateway.localhost:3000/
- http://docs-ipfs-tech.ipns.helia-sw-gateway.localhost:3000/

It allows us to have release artifacts which work everywhere.

## Notes & open questions

Long term, we should make `@helia/verified-fetch`  support `http(s)://` URLs for path, subdomain, and dnslink gateways, removing the need for creating `ipfs://` and `ipns://` URLs.

## Change checklist


src/lib/dns-link-labels.ts
```diff
@@ -1,13 +1,35 @@
...
+
 /**
- * We can receive either a peerId string or dnsLink label string here. PeerId strings do not have dots or dashes.
+ * Checks if label looks like inlined DNSLink.
+ * (https://specs.ipfs.tech/http-gateways/subdomain-gateway/#host-request-header)
  */
-export function isDnsLabel (label: string): boolean {
-  return ['-', '.'].some((char) => label.includes(char))
+export function isInlinedDnsLink (label: string): boolean {
+  return isValidDnsLabel(label) && label.includes('-') && !label.includes('.')
```
lidel, line 32: :information_source: to avoid false-positives added code to ensure an inlined DNSLink
- does not parse as a valid CID
- does not include `.`
- includes at least one `-`

src/lib/heliaFetch.ts
```diff
@@ -1,14 +1,13 @@
...
-import { dnsLinkLabelDecoder, isDnsLabel } from './dns-link-labels.ts'
 import type { Helia } from '@helia/interface'
 
 export interface HeliaFetchOptions {
   path: string
   helia: Helia
   signal?: AbortSignal
   headers?: Headers
-  origin?: string | null
+  id?: string | null
```
lidel, line 10: :information_source:  had to rename, would be confusing for web devs  ("origin" in web security context, including `window.location`, means at minimum both protocol and the authority identifier).

src/lib/path-or-subdomain.ts
lidel, line 1: :thought_balloon: I did not clean this up to make the PR smaller and easier to review.

My plan is to move majority of subdomain handling to `@helia/verified-fetch` in future PRs and then get back here and decide what minimal handling makes sense to remain in this project.

src/sw.ts
```diff
@@ -91,23 +91,34 @@ const isRootRequestForContent = (event: FetchEvent): boolean => {
   return isRootRequest // && getCidFromUrl(event.request.url) != null
 }
 
-function getSubdomainParts (request: Request): { origin: string | null, protocol: string | null } {
+function getSubdomainParts (request: Request): { id: string | null, protocol: string | null } {
   const urlString = request.url
-  const url = new URL(urlString)
-  const subdomain = url.hostname.replace(`.${BASE_URL}`, '')
-  const subdomainRegex = /^(?<origin>[^/]+)\.(?<protocol>ip[fn]s)?$/
```
lidel, line 98: :information_source:  this produced invalid results for `docs.ipfs.tech.ipns.example.com`, the new code walks labels from right to left, which is better order for avoiding false-positive matches.

lidel (commented):
Some highlights below.

public/_redirects
lidel, line 1: :thought_balloon:  I was not able to find where  `?helia-sw` is handled  – assumed it is not wired up yet,  but I've updated `_redirects` so it will work on web (subdomain or dnslink) gateway once we add it.

@SgtPooki what was the plan around `?helia-sw-subdomain` and `?helia-sw`?  I want to add origin isolation enforcement in follow-up PR and most likely will have to touch this.

SgtPooki: ?helia-sw & ?helia-sw-subdomain are no longer required and should be able to be removed

SgtPooki: they were used to help with registering the service worker originally, but now, if the service worker is hit, we never render the jsx, and if the service worker isn't hit, we determine (in app.tsx) if we need to load the <redirectPage> or not

src/lib/heliaFetch.ts
```diff
@@ -116,29 +115,26 @@ function changeCssFontPath (path: string): string {
  * heliaFetch should have zero awareness of whether it's being used inside a service worker or not.
  *
  * The `path` supplied should be either:
- * * /ipfs/CID
- * * /ipns/DNSLink
- * * /ipns/key
+ * * /ipfs/CID      (https://docs.ipfs.tech/concepts/content-addressing/)
+ * * /ipns/DNSLink  (https://dnslink.dev/)
+ * * /ipns/IPNSName (https://specs.ipfs.tech/ipns/ipns-record/#ipns-name)
```
SgtPooki, line 120: +1

src/sw.ts
SgtPooki: should we redirect from subdomains like "docs.ipfs.tech" to "docs-ipfs-tech" ?

src/index.tsx
```diff
@@ -15,13 +14,14 @@ const sw = await navigator.serviceWorker.register(new URL('sw.ts', import.meta.u
...
 // SW did not trigger for this request
-if (isPathOrSubdomainRequest(BASE_URL, window.location)) {
+if (isPathOrSubdomainRequest(window.location)) {
   // but the requested path is something it should, so show redirect and redirect to the same URL
   root.render(
     <RedirectPage />
   )
   window.location.replace(window.location.href)
 } else {
+  // TODO: add detection of DNSLink gateways (alowing use with Host: en.wikipedia-on-ipfs.org)
```
SgtPooki, line 24: if `en.wikipedia-on-ipfs.org.ipns.sw-host.tld` is requested, this will never be hit.

public/_redirects
```diff
@@ -1,2 +1 @@
-/ipns/* /?helia-sw=/ipns/:splat 302
-/ipfs/* /?helia-sw=/ipfs/:splat 302
+/* /?helia-sw=/:splat 302
```
SgtPooki, line 1: we should be able to remove this and just resolve https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/28

SgtPooki (commented):
some comments and questions

SgtPooki (approved):

PR #56, opened 2024-02-27
feat: support ens resolution

### LINKED ISSUE #23: Support the same  DNSLink TLDs as Kubo and Rainbow
### Summary

By default, same TLDs should work in WS as in Kubo/Rainbow.

We want service worker gateway to be a drop-in option for people who currently load simple websites via Kubo or Rainbow.

### Background

Kubo uses DoH at `resolver.cloudflare-eth.com/dns-query` for `.eth.` and `.crypto` names.

Unfortunately the endpoint does not have liberal CORS set up, and we use `https://dns.eth.limo/dns-query` for `.eth` instead. This means `.crypto` is broken, because it has no DNSLink resolver set up.

### Solutions

Our options are:
- [ ] ~quick/short-term: set up https://github.com/ipshipyard/waterworks-community/issues/6 and switch to it~
- [x] long-term: have Cloudflare fix CORS at their endpoint, and switch back to it (we will discuss timeline this week)
  - fixed, https://github.com/ipshipyard/waterworks-community/issues/6#issuecomment-1964289603

SgtPooki:
- chore: use latest verified-fetch RC
- feat: add cloudflare eth resolver

## Description

Adds cloudflare ens resolver (which falls back to regular dns resolution) and updates verified-fetch to the latest RC.

fixes https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/23

## Notes & open questions

**SHOULD NOT BE MERGED** because cloudflare is still failing CORS. See https://github.com/ipshipyard/waterworks-community/issues/6#issuecomment-1965703554

## Change checklist


src/lib/heliaFetch.ts
```diff
@@ -126,7 +129,12 @@ function changeCssFontPath (path: string): string {
  *
  */
 export async function heliaFetch ({ path, helia, signal, headers, id, protocol }: HeliaFetchOptions): Promise<Response> {
```
SgtPooki, line 126: we should remove `helia` here and the `get-helia` since we can't pass both custom dnsResolvers and a custom helia instance.

SgtPooki (commented):
self review

SgtPooki:
ens resolution is working with the last RC version of verified-fetch and the config from https://github.com/ipshipyard/waterworks-infra/pull/30 deployed to `sgtpooki.com/dns-query` that is currently returning 404 at delegated-ipfs.dev/dns-query.

SgtPooki (commented):
self review

Issue #30, opened 2024-02-23
Origin isolation warnings, tests and enforcement


lidel:
> Extracted from https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/20#issuecomment-1955188119

### For `inbrowser.link` 

The goal is to provide a subdomain gateway for loading websites in web browser.

Lack of isolation is a security bug. Subdomain mode MUST be enforced.

### For General Availability

Good security is like water, if things are right, fishes should not even notice it. 

For end users:
- We need to protect end user from subdomain misconfigurations, and at least warn them when SW gateway is deployed without proper origin isolation.
  - SW code must be smart enough to do the right thing when deployed on third-party domains, without additional configuration.  
  - No shortcuts, no hardcoding "subdomain hostnames". This has to be automated, and work with any third-party domain.

For gateway Operators: 
- Setup complexity should be limited to getting wildcard TLS certs. 
  - Operators should know what to do/fix by just opening website via SW gateway and reading the same error as end user.

## Initial design 

- Before SW is handling responses, run smoke-tests for origin isolation setup
- IF subdomain setup is not available
  - Instead of loading screen, produce a loud warning to the user 
    - UI should be similar to TLS or other [security errors in browsers](https://support.mozilla.org/en-US/kb/what-does-your-connection-is-not-secure-mean) or disclaimers at [public-gateway-checker](https://ipfs.github.io/public-gateway-checker) or [addressing docs](https://docs.ipfs.tech/how-to/address-ipfs-on-web), 
    - Inform user that web features such as _redirects and Origin isolation will be missing
    - User can continue with web browsing via path gateway, but they need to explicitly click "i accept the risk" button. Path gateway should always p
- detects when we are on example.com/ipfs or /ipns path _while_ subdomains for [subdomain gateway](https://specs.ipfs.tech/http-gateways/subdomain-gateway/) are present (`*.ipfs.example.com` works)
  - Investigate `fetch` API, as noted in https://github.com/ipfs/helia/pull/439#pullrequestreview-1891751926
- returns HTTP 301  subdomain redirect ([example.com/ipfs/cid → cid.ipfs.example.com](https://specs.ipfs.tech/http-gateways/subdomain-gateway/#migrating-from-path-to-subdomain-gateway)) to enforce origin isolation.  
  - `window.location.replace` may be even better as it removes path URL from browser history.
  
 
## Tasks

Before ETH Denver:
[...]

PR #24, opened 2024-02-21
feat: create config page for sw settings

SgtPooki:
partially addresses #22

```[tasklist]
### Todo
- [x] make sure gateway and router config is being registered with helia instance properly on subdomains
- [ ] center gear icon padding
- [x] add auto-reload config
```

PR #60, opened 2024-02-28
fix: enforce origin isolation on subdomain gws

lidel:
## Description

Towards #30: this PR detects when a path gateway URL is loaded  on host that supports subdomain gateway URLs, and redirects to unique origin.

To test:

1. Open http://helia-sw-gateway.localhost:3000 to register SW
1. Open http://helia-sw-gateway.localhost:3000/ipns/cid.ipfs.tech 
1. Page should redirect  to http://cid-ipfs-tech.ipns.helia-sw-gateway.localhost:3000

Caveats: 
- in Brave this works only with `helia-sw-gateway.localhost`, they seem to block `localhost`, so use `helia-sw-gateway.localhost` for testing
- check depends on CORS

## Notes & open questions

- [x] added follow-up tasks to  #30

## Change checklist

### DIFF
--- a/src/components/config.tsx
+++ b/src/components/config.tsx
@@ -44,6 +44,7 @@ export default (): JSX.Element | null => {
       return
     }
     // we get the iframe origin from a query parameter called 'origin', if this is loaded in an iframe
+    // TODO: why we need this origin here? where is targetOrigin used?
     const targetOrigin = decodeURIComponent(window.location.search.split('origin=')[1])
     const config = await getConfig()
 
--- a/src/lib/path-or-subdomain.ts
+++ b/src/lib/path-or-subdomain.ts
@@ -1,14 +1,84 @@
+import { base32 } from 'multiformats/bases/base32'
+import { base36 } from 'multiformats/bases/base36'
+import { CID } from 'multiformats/cid'
+import { dnsLinkLabelEncoder } from './dns-link-labels.ts'
+
 // TODO: dry, this is same regex code  as in  getSubdomainParts
 const subdomainRegex = /^(?<id>[^/]+)\.(?<protocol>ip[fn]s)\.[^/]+$/
 const pathRegex = /^\/(?<protocol>ip[fn]s)\/(?<path>.*)$/
 
-export const isPathOrSubdomainRequest = (location: Pick<Location, 'hostname' | 'pathname'>): boolean => {
-  const subdomain = location.hostname
-  const subdomainMatch = subdomain.match(subdomainRegex)
+export const isPathOrSubdomainRequest = (location: Pick<Location, 'host' | 'pathname'>): boolean => {
+  return isPathGatewayRequest(location) || isSubdomainGatewayRequest(location)
+}
 
+export const isSubdomainGatewayRequest = (location: Pick<Location, 'host' | 'pathname'>): boolean => {
+  const subdomainMatch = location.host.match(subdomainRegex)
+  return subdomainMatch?.groups != null
+}
+
+export const isPathGatewayRequest = (location: Pick<Location, 'host' | 'pathname'>): boolean => {
   const pathMatch = location.pathname.match(pathRegex)
-  const isPathBasedRequest = pathMatch?.groups != null
-  const isSubdomainRequest = subdomainMatch?.groups != null
+  return pathMatch?.groups != null
+}
 
-  return isPathBasedRequest || isSubdomainRequest
+/**
+ * Origin isolation check and enforcement
+ * https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/30
+ */
+export const findOriginIsolationRedirect = async (location: Pick<Location, 'protocol' | 'host' | 'pathname' | 'search' | 'hash' >): Promise<string | null> => {
+  if (isPathGatewayRequest(location) && !isSubdomainGatewayRequest(location)) {
+    const redirect = await isSubdomainIsolationSupported(location)
+    if (redirect) {
+      return toSubdomainRequest(location)
+    }
+  }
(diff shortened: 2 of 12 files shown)


src/lib/path-or-subdomain.ts
```diff
@@ -1,14 +1,84 @@
...
+const isSubdomainIsolationSupported = async (location: Pick<Location, 'protocol' | 'host' | 'pathname'>): Promise<boolean> => {
+  // TODO: do this test once and expose it as cookie / config flag somehow
+  const testUrl = `${location.protocol}//bafkqaaa.ipfs.${location.host}`
+  try {
+    const response: Response = await fetch(testUrl)
+    return response.status === 200
+  } catch (_) {
+    return false
+  }
+}
```
lidel, line 47: :thought_balloon: this is not a big deal because it only happens once when on /ipfs/cid paths, is not executed when on subdomain host, but perhaps we could leverage config passing setup based on iframe introduced in #24

src/components/config.tsx
```diff
@@ -44,6 +44,7 @@ export default (): JSX.Element | null => {
       return
     }
     // we get the iframe origin from a query parameter called 'origin', if this is loaded in an iframe
+    // TODO: why we need this origin here? where is targetOrigin used?
     const targetOrigin = decodeURIComponent(window.location.search.split('origin=')[1])
```
lidel, line 48: @SgtPooki is this something we want to keep, or can it be removed?
iirc if we don't have query param, we would reuse cache, making subdomain page loads faster

lidel:
Merging this PR will force redirect to subdomains when available and liberal CORS headers are present. This includes `inbrowser.dev`.

Given how sensitive the lack of isolation is for websites atm, I'm thinking about merging this as-is, to secure `inbrowser.dev`, and adding regression tests and optimizations in follow-up PRs. We should not block critical security fix on these.

Ok to proceed @SgtPooki @aschmahmann ?

src/components/config.tsx
SgtPooki: This is only for loading config in iframe. Iframe needs to know parent origin to postmessage.

Target origin (main window) is subdomain, iframe is root domain.

SgtPooki: So if we can do that another way, we could remove this

src/lib/path-or-subdomain.ts
```diff
@@ -1,14 +1,84 @@
...
+  if (isPathGatewayRequest(location) && !isSubdomainGatewayRequest(location)) {
+    const redirect = await isSubdomainIsolationSupported(location)
+    if (redirect) {
+      return toSubdomainRequest(location)
+    }
+  }
+  return null
+}
+
+const isSubdomainIsolationSupported = async (location: Pick<Location, 'protocol' | 'host' | 'pathname'>): Promise<boolean> => {
```
SgtPooki, line 38: its worth adding a comment here saying that for our SW usecase, we're only worried about checking for a 200 response and no TLS error.

src/lib/path-or-subdomain.ts
```diff
@@ -1,14 +1,84 @@
...
+    switch (ns) {
+      case 'ipfs':
+        // Base32 is case-insensitive and allows CID with popular hashes like sha2-256 to fit in a single DNS label
+        id = CID.parse(id).toV1().toString(base32)
+        break
+      case 'ipns':
+        // IPNS Names are represented as Base36 CIDv1 with libp2p-key codec
+        // https://specs.ipfs.tech/ipns/ipns-record/#ipns-name
+        // eslint-disable-next-line no-case-declarations
+        const ipnsName = CID.parse(id).toV1()
```
SgtPooki, line 67: why are we `CID.parse` ing a dnslink/peerid?

src/redirectPage.tsx
```diff
@@ -22,6 +23,19 @@ export default function RedirectPage (): JSX.Element {
...
+      // enforce  early when loaded before SW was registered
+      const originRedirect = await findOriginIsolationRedirect(window.location)
+      if (originRedirect !== null) {
+        window.location.replace(originRedirect)
+      }
+    }
+    void originEnforcement()
+  }, [])
+  */
+
```
SgtPooki:
```suggestion
```

src/redirectPage.tsx
```diff
@@ -5,6 +5,7 @@ import { HeliaServiceWorkerCommsChannel } from './lib/channel.ts'
 import { setConfig, type ConfigDb } from './lib/config-db.ts'
 import { getSubdomainParts } from './lib/get-subdomain-parts'
 import { error } from './lib/logger.ts'
+// import { findOriginIsolationRedirect } from './lib/path-or-subdomain.ts'
```
SgtPooki:
```suggestion
```

SgtPooki (approved):
a few comments / nits, but overall fine with getting this in

SgtPooki:
redirect to /wiki for http://helia-sw-gateway.localhost/ipns/en.wikipedia-on-ipfs.org doesn't work, but other than that, this seems to work well

src/lib/path-or-subdomain.ts
lidel: Valid CIDv1 in base32 will be also a valid DNS name.
So we parse string as CID first, and ONLY assume DNS if CID parsing failed.

If we did it the other way, ISP or DNS operator could return spoofed DNS name that is also a valid CID, and do MITM on users who only use cryptographic identifies.

src/components/config.tsx
lidel: Could we use `#hash` instead? It does not get sent to server, but JS in inframe should see it.

PR #76, opened 2024-03-01
fix: load the iframe in the same port as the parent domain

2color:
## Background

This makes sure that the config iframe is loaded from the same port. Otherwise this fails when not on port 80, e.g. using the local development server. 

This also makes it possible to use subdomain resolution locally with the Webpack dev server.

## Change checklist

### DIFF
--- a/src/redirectPage.tsx
+++ b/src/redirectPage.tsx
@@ -9,7 +9,7 @@ import { error } from './lib/logger.ts'
 const ConfigIframe = (): JSX.Element => {
   const { parentDomain } = getSubdomainParts(window.location.href)
 
-  const iframeSrc = `${window.location.protocol}//${parentDomain}/config?origin=${encodeURIComponent(window.location.origin)}`
+  const iframeSrc = `${window.location.protocol}//${parentDomain}:${window.location.port}/config?origin=${encodeURIComponent(window.location.origin)}`
 
   return (
     <iframe id="redirect-config-iframe" src={iframeSrc} style={{ width: '100vw', height: '100vh', border: 'none' }} />
--- a/src/redirectPage.tsx
+++ b/src/redirectPage.tsx
@@ -9,7 +9,8 @@ import { error } from './lib/logger.ts'
 const ConfigIframe = (): JSX.Element => {
   const { parentDomain } = getSubdomainParts(window.location.href)
 
-  const iframeSrc = `${window.location.protocol}//${parentDomain}:${window.location.port}/config?origin=${encodeURIComponent(window.location.origin)}`
+  const portString = window.location.port === '' ? '' : `:${window.location.port}`
+  const iframeSrc = `${window.location.protocol}//${parentDomain}${portString}/config?origin=${encodeURIComponent(window.location.origin)}`
 
   return (
     <iframe id="redirect-config-iframe" src={iframeSrc} style={{ width: '100vw', height: '100vh', border: 'none' }} />


SgtPooki:
> This also makes it possible to use subdomain resolution locally with the Webpack dev server.

🔥

src/redirectPage.tsx
```diff
@@ -9,7 +9,7 @@ import { error } from './lib/logger.ts'
 const ConfigIframe = (): JSX.Element => {
   const { parentDomain } = getSubdomainParts(window.location.href)
 
-  const iframeSrc = `${window.location.protocol}//${parentDomain}/config?origin=${encodeURIComponent(window.location.origin)}`
+  const iframeSrc = `${window.location.protocol}//${parentDomain}:${window.location.port}/config?origin=${encodeURIComponent(window.location.origin)}`
```
SgtPooki: this will be incorrect when no port is specified. i.e. `http://localhost:/config?...` or even `https://inbrowser.link:/config?`. Apparently it's not breaking, but still:

```suggestion
  const portString = window.location.port === '' ? '' : `:${window.location.port}`
  const iframeSrc = `${window.location.protocol}//${parentDomain}${portString}/config?origin=${encodeURIComponent(window.location.origin)}`
```

SgtPooki (changes requested):

SgtPooki:
FYI, even with this fix, trying to load the iframe is failing in brave for me on localhost. no errors emitted but looking in the iframe via element inspector shows:

```
var loadTimeDataRaw = {
  "details": "Details",
  "errorCode": "ERR_BLOCKED_BY_CLIENT",
  "fontfamily": "system-ui, sans-serif",
  "fontfamilyMd": "system-ui, sans-serif",
  "fontsize": "75%",
  "heading": {
    "hostName": "localhost",
    "msg": "\u003Cspan jscontent=\"hostName\">\u003C/span> is blocked"
  },
  "hideDetails": "Hide details",
  "iconClass": "icon-generic",
  "language": "en",
  "reloadButton": {
    "msg": "Reload",
    "reloadUrl": "http://localhost:3000/config?origin=http%3A%2F%2Fblog-libp2p-io.ipns.localhost%3A3000"
  },
  "suggestionsDetails": [],
  "suggestionsSummaryList": [],
  "summary": {
    "failedUrl": "http://localhost:3000/config?origin=http%3A%2F%2Fblog-libp2p-io.ipns.localhost%3A3000",
    "hostName": "localhost",
    "msg": "This page has been blocked by Brave"
  },
  "textdirection": "ltr",
  "title": "localhost"
};
```

image

2color:
> FYI, even with this fix, trying to load the iframe is failing in brave for me on localhost. no errors emitted but looking in the iframe via element inspector shows:

Yep. I was also able to reproduce this: 
Screenshot 2024-03-04 at 2 29 47 pm

It's because of [Brave Shields](https://brave.com/shields/) which prevents tracking on sites which is commonly done via iframes of a different origin, as is the case here (since the iframe is loaded from the root origin which is different to the subdomain gateway origin).

Disabling Brave Shields resolves this, but I think it may be worth rethinking whether we need an iframe for this (as I asked in https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/75).

src/redirectPage.tsx
2color: Nice! Thanks

PR #77, opened 2024-03-01
feat: improve styling of header

2color:
## What's in this PR

- Use flex space-between to align the header items
- Add text to header saying that this is an **IPFS Service Worker Gateway**

## Change checklist


SgtPooki (approved):
looks great!

Issue #62, opened 2024-02-28
Trailing / for UnixFS directory normalization is missing 


achingbrain:
Trying to click a footnote takes you back to the Wikipedia homepage instead of the footer of the current page:

<img width="727" alt="image" src="https://github.com/ipfs-shipyard/helia-service-worker-gateway/assets/665810/a02bcad8-5674-4f2f-b981-d0ffad62e325">

This is because the footnote href is a relative link up one directory:

<img width="414" alt="image" src="https://github.com/ipfs-shipyard/helia-service-worker-gateway/assets/665810/08fe2e6d-080b-47f2-89f4-6226d9279a35">

All local links seem to have this problem:

<img width="483" alt="image" src="https://github.com/ipfs-shipyard/helia-service-worker-gateway/assets/665810/09bf67de-0c46-4754-910a-7275449997fa">

lidel:
@achingbrain do you have specific article to reproduce?  Check if this was also broken when loaded via dweb.link gateway.

iirc this ancient wikipedia mirror has JS that fixes relative links post-load, so if that JS file failed to load via SW, links will be broken.

achingbrain:
Sure, I posted a bunch in https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/061

E.g. click any linked text here: https://bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze.ipfs.sw.sgtpooki.com/wiki/Archaeology

> iirc this ancient wikipedia mirror has JS that fixes relative links post-load

Ah, ok - there are some js errors in the console - a syntax error in the file, maybe that's it.  Is there a later `CID` for the mirror?

lidel:
Ok, I know what is the problem.
Yet another signal we should set up https://github.com/ipfs/gateway-conformance  with proper coverage.

SW gateway does not normalize directory URLs. They must end with `/`. Gateway ensure that by redirecting UnixFS directory paths to URL that ends with trailing `/`.

Example:
https://bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze.ipfs.dweb.link/wiki/Archaeology returns HTTP 302 to https://bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze.ipfs.dweb.link/wiki/Archaeology/ and that makes relative links work correctly.

[...]

PR #82, opened 2024-03-04
chore: verified-fetch v1.1.0

2color:
## Description

This PR should help with https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/62 by upgrading to the latest version of helia/verified-fetch. 

## Change checklist


lidel (approved):
This turned out to be not enough to fix #62 (see details in https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/62#issuecomment-1977661456), but version bump sgtm.

Issue #73, opened 2024-02-29
Cache final Responses


lidel:
afaik `Response` produced by service worker is NOT cached by the browser, unless it is explicitly put in the cache. Right now we don't do that, so page reload trigggers fetch of all chunks as blocks, which are returned from cache and re-assembled again, rather than the final result being returned from cache.

If we add final Responses to  cache, page reload will not trigger block requests for resources which were already handled.

Ref. 

- https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers
- https://developer.chrome.com/docs/workbox/caching-strategies-overview is a good resource as it describes caching strategies we want to implement:
  -  /ipfs/: these are immutable, we want "cache first, falling back to network" strategy
  - /ipns/: these are mutable, and due to this more expensive. I believe we want "Stale-while-revalidate" strategy as it will make all websites feel snappy, work in offline mode and when censored, and leverage SW powers the most.

SgtPooki:
we should be caching only 2xx,3xx permanently (for IPFS only). maybe 15s for errors.

IPNS is being cached at verified-fetch layer (IIRC)

2color:
I've been investigating this and what the advantages of using the cache API would be over IndexedDB (which is already implemented but not used)

The following is a useful resource https://medium.com/dev-channel/offline-storage-for-progressive-web-apps-70d52695513c#.19w8r1c4o

Some initial insights:
- It seems that both IndexedDB and the Cache API share the same storage quota
- By using the indexedDB block store, we may benefit from more granular block level caching
- "Items in a Cache do not get updated unless explicitly requested; they don't expire unless deleted." The same applies to indexedDB

@lidel Are there any specific benefits of the Cache API over indexedDB you are aware of?

lidel:
@2color Thanks for looking into this. Indeed, shared quota is a bummer, but let's remember it applies only to Cache API, not individual block responses. 

Cache API is still something we need. Cache API stores final bytes produced by IPFS stack, which means work necessary for turning IPFS DAG into final file occurs only once, and not on every page load for every image etc on the page.

Block level caching is useful when loading new subpath on a path we already visited, but iiuc we already get that from trustless-gateway.link responses being cached (responses have `cache-control` header that tells browser to  cache each block for as long as possible), so lower priority.

I think the priority is to set up Cache API, especially for `/ipns/..` content path responses, and then  fine to look into IndexedDB-based blockstore and see if it improves any metric.

[...]

PR #89, opened 2024-03-06
refactor: remove unused helia, use verified fetch, and simplify code

### LINKED ISSUE #88: Slow iframe requests to root domain (unintended cache bypass)
Right now, every subdomain load makes additional request to the root domain with custom query parameter, which is effectively bypassing caches (even one from https://github.com/ipshipyard/waterworks-infra/pull/42) and forces rainbow to do work every time again.

This is why we sometimes see `/config?origin` being stuck in pending state:

> 

And then often resulting in 504 for the iframe:
> 

### Solution

We should lean towards maximizing HTTP and browser cache and switch config logic from server-side `/path?query` to client-side `#framents`

- iframe should make request to `/#config` of the root domain, not `/config`. This way already cached `/` response can be returned instead of bypassing caches and rainbow having to produce it again
- `?origin` query param should be replaced with frament param, e.g. `/#config@origin=`. `?origin` will create separate cache entry per origin. `/#foo` is always better, works without SW being in place, URL hash fragments are not sent to HTTP server, allowing for cache reuse, but can still be read by JS running in iframe

2color:
## PR Summary

This PR is mostly refactoring with some small changes to make it easier to implement caching. 

- Replace the singleton helia instance in favour of a verifiedFetch instance because the helia instance wasn't actually used
- Remove the indexedDB block store - this is a temporary measure until I decide how to implement the cache in #73. Since the helia instance wasn't used and verifiedFetch was instantiated on every request, this has no effect
- Simplify the heliaFetch function and break it into more easily testable functions 
- Move the default routers and gateways config to storage so that the user can also customise the default config (not finished since it's initially loaded from localStorage which can probably removed [as per discussion](https://github.com/ipfs-shipyard/helia-service-worker-gateway/discussions/86#discussioncomment-8678731))
- Fixes #88 by loading the config page using a hash fragment instead of query param
- 

## Notes & open questions
- We may still want to use indexedDB blockstore for caching (related #73) . If so, see https://github.com/ipfs/helia-verified-fetch/pull/13 which should enable us to use it while customising the DoH endpoints. 
- Config passing between origins needs to be more carefully tested. It seems like in some cases this is not working properly, e.g. when auto-reload is set up and there isn't enough time for the message to be passed from the root to the subdomain origin. 

## Change checklist


SgtPooki:
A few comments before I review:

> Replace the singleton helia instance in favour of a verifiedFetch instance because the helia instance wasn't actually used

:+1: great, yea this needed done!

> Move the default routers and gateways config to storage so that the user can also customise the default config (not finished since it's initially loaded from localStorage which can probably removed https://github.com/ipfs-shipyard/helia-service-worker-gateway/discussions/86#discussioncomment-8678731)

~~moved them to which storage if we removed indexedDB?~~ nvm, i see that they're moved to localStorage.

> Move the default routers and gateways config to storage so that the user can also customise the default config (not finished since it's initially loaded from localStorage which can probably removed https://github.com/ipfs-shipyard/helia-service-worker-gateway/discussions/86#discussioncomment-8678731)

I believe in the last helia WG call we discussed not removing localStorage.

> We may still want to use indexedDB blockstore for caching (related https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/73) . If so, see https://github.com/ipfs/helia-verified-fetch/pull/13 which should enable us to use it while customising the DoH endpoints.

~~We do still want to use indexedDB for passing data to and from the config page.~~ I don't think we need the indexedDB blockstore at all. A plain indexedDB instance should work

> Config passing between origins needs to be more carefully tested. It seems like in some cases this is not working properly, e.g. when auto-reload is set up and there isn't enough time for the message to be passed from the root to the subdomain origin.

There is a TODO to use the commsChannel to use passMessageAndWaitForResponse (or whatever its called) which would require waiting until the message is passed and sw responds back

src/components/config.tsx
```diff
@@ -45,9 +45,10 @@ export default (): JSX.Element | null => {
     }
     // we get the iframe origin from a query parameter called 'origin', if this is loaded in an iframe
     // TODO: why we need this origin here? where is targetOrigin used?
-    const targetOrigin = decodeURIComponent(window.location.search.split('origin=')[1])
+    const targetOrigin = decodeURIComponent(window.location.hash.split('@origin=')[1])
     const config = await getConfig()
-
+    // eslint-disable-next-line no-console
+    console.log('config-page: postMessage config to origin ', config, origin)
```
SgtPooki:
```suggestion
```

src/get-helia.ts
```diff
@@ -1,30 +1,20 @@
-import { trustlessGateway } from '@helia/block-brokers'
-import { createHeliaHTTP } from '@helia/http'
-import { delegatedHTTPRouting } from '@helia/routers'
-import { IDBBlockstore } from 'blockstore-idb'
-import { IDBDatastore } from 'datastore-idb'
+import { dnsJsonOverHttps } from '@helia/ipns/dns-resolvers'
```
SgtPooki, line 1: needs a file-name change too.

src/lib/heliaFetch.ts
```diff
@@ -123,45 +73,42 @@ function changeCssFontPath (path: string): string {
...
-    }
-    const pathRootString = pathParts[pathPartIndex++]
-    const contentPath = pathParts.slice(pathPartIndex++).join('/')
-    verifiedFetchUrl = `${namespaceString}://${pathRootString}/${changeCssFontPath(contentPath)}`
-  }
-
-  // eslint-disable-next-line no-console
-  console.log('verifiedFetch for ', verifiedFetchUrl)
-  return verifiedFetch(verifiedFetchUrl, {
+export async function heliaFetch ({ verifiedFetch, verifiedFetchUrl, signal, headers }: HeliaFetchOptions): Promise<Response> {
```
SgtPooki: we could probably remove this method entirely while we're at it, and call `verifiedFetch` directly from the service worker

src/sw.ts
```diff
@@ -76,7 +79,10 @@ const fetchHandler = async ({ path, request }: FetchHandlerArg): Promise<Respons
   const abortController = AbortSignal.timeout(5 * 60 * 1000)
   try {
     const { id, protocol } = getSubdomainParts(request.url)
-    return await heliaFetch({ path, helia, signal: abortController, headers: request.headers, id, protocol })
+    const verifiedFetchUrl = getVerifiedFetchUrl({ id, protocol, path })
+    log('verifiedFetch for ', verifiedFetchUrl)
+
+    return await heliaFetch({ verifiedFetch, verifiedFetchUrl, signal: abortController, headers: request.headers })
```
```suggestion
    return await verifiedFetch(verifiedFetchUrl, {
      signal,
      headers,
      // TODO redirect: 'manual', // enable when http urls are supported by verified-fetch: https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/62#issuecomment-1977661456
      onProgress: (e) => {
        trace(`${e.type}: `, e.detail)
      }
    })
```

src/sw.ts
```diff
@@ -97,6 +103,7 @@ const fetchHandler = async ({ path, request }: FetchHandlerArg): Promise<Respons
     return new Response('heliaFetch error: ' + errorMessage, { status: 500 })
```
SgtPooki, line 207:
```suggestion
    return new Response('verifiedFetch error: ' + errorMessage, { status: 500 })
```

src/sw.ts
```diff
@@ -1,28 +1,35 @@
...
+
   channel.onmessagefrom('WINDOW', async (message: MessageEvent<ChannelMessage<'WINDOW', null>>) => {
     const { action } = message.data
     switch (action) {
       case 'RELOAD_CONFIG':
-        void getHelia().then((newHelia) => {
-          helia = newHelia
+        void getVerifiedFetch().then((newVerifiedFetch) => {
+          verifiedFetch = newVerifiedFetch
         })
```
SgtPooki: this is already async.. i wonder why I didn't use `verifiedFetch = await ...`

this should be updated

SgtPooki (approved):
I'm a fan of these changes in general. I think there are a few changes we should get in before merging though. 

Thanks Daniel!

src/sw.ts
SgtPooki: fixed in https://github.com/ipfs-shipyard/helia-service-worker-gateway/pull/94

src/sw.ts
SgtPooki: fixed in https://github.com/ipfs-shipyard/helia-service-worker-gateway/pull/94

src/components/config.tsx
SgtPooki: fixed in https://github.com/ipfs-shipyard/helia-service-worker-gateway/pull/94

src/get-helia.ts
SgtPooki: fixed in https://github.com/ipfs-shipyard/helia-service-worker-gateway/pull/94

SgtPooki: Fixed in #95

src/lib/heliaFetch.ts
SgtPooki: Fixed in #95

src/sw.ts
SgtPooki: Fixed in #95

SgtPooki: I mean #95

PR #92, opened 2024-03-07
feat: add caching to service worker

### LINKED ISSUE #73: Cache final Responses

2color:
## Description

This PR adds caching for responses using the [Cache API](https://developer.mozilla.org/en-US/docs/Web/API/Cache).

Fixes #73 

## Notes & open questions

- Do we want to also cache for the `isRootRequestForContent` case?: https://github.com/ipfs-shipyard/helia-service-worker-gateway/blob/3cc5b9882dbbf347a8c5fd51df311c8705dc6fc8/src/sw.ts#L94-L96 
- Right now the cache key is based on the `event.request`. This seems reasonable but maybe it's not optimal?
  - Update: it's not possible to use the IPFS URL for cache. Hence the use of the request 

## TODO
- Double check how the caching and invalidation works for IPNS paths 

## Change checklist


SgtPooki:
> Right now the cache key is based on the event.request. This seems reasonable but maybe it's not optimal?

We should probably use the verifiedFetchUrl you have in the other PR

2color:
> > Right now the cache key is based on the event.request. This seems reasonable but maybe it's not optimal?
> 
> We should probably use the verifiedFetchUrl you have in the other PR

The Cache API does not support URLs with the `ipfs://` protocol scheme: 
Screenshot 2024-03-11 at 1 45 39 pm

src/sw.ts
```diff
@@ -94,7 +95,32 @@ self.addEventListener('fetch', (event) => {
     // intercept and do our own stuff...
     event.respondWith(fetchHandler({ path: url.pathname, request }))
   } else if (isSubdomainRequest(event)) {
-    event.respondWith(fetchHandler({ path: url.pathname, request }))
+    const cacheKey = event.request.url
```
lidel: nit: cacheKey should include `Accept` header from request (if present), to avoid future bugs where CAR is returned instead of HTML or vice versa.

src/sw.ts
```diff
@@ -94,7 +95,32 @@ self.addEventListener('fetch', (event) => {
...
+    log(`helia-sw: cache key:`, cacheKey.toString())
+
+    event.respondWith((async () => {
+      const cache = await caches.open(IPFS_CACHE)
```
lidel: :thought_balloon:  if we had separate cache instances for mutable and immutable responses, it will be easier to implement separate cache eviction strategies for each.

src/sw.ts
```diff
@@ -94,7 +95,32 @@ self.addEventListener('fetch', (event) => {
...
+      }
+
+      // 👇 fetch because no cached response was found
+      const response = await fetchHandler({ path: url.pathname, request })
+
+      if(response.ok) {
+        // 👇 only cache successful responses
+        log('helia-ws: storing cache key %s in cache', cacheKey)
+        // Clone the response since streams can only be consumed once.
+        cache.put(cacheKey, response.clone())
```
lidel: :thought_balloon:  this works for immutable `/ipfs/cid`, but to land this PR we need separate logic for mutable responses, even if it is very simple, such as "cache for 1h, then update from network". I think the end goal is for verified-fetch to return Response with meaningful `Cache-Control` and our SW here just acting on `max-age` and `Expiration` from that, so we don't have to implement perfect caching strategy here.

One way of implementing expiration on `/ipns/..` is to add `Expires` header to a cached `/ipns/..` response (everything other than immutable path or subdomain), and then when it is read back via `cache.match` check if expiration happenend. If it expired, fetch new one from network. 

We can switch it to stale-while-revalidate later by always returning cached result and doing update async.

lidel (changes requested):

src/sw.ts
2color: For sites deployed to IPFS, requests will either be mutable or immutable. Can you give an example where some request/response pairs with be mixed (both mutable and immutable)? 

For example, `vitalik-eth.ipns.inbrowser.link` will result in all of its subresources requested as a path on the root IPNS names, i.e. `ipns://vitalik.eth/....`. Given that, it means that we need to treat all of them as mutable.

src/sw.ts
2color: In theory we could simplify it here a bit and just use the request object rather than a custom cache key. The cache API looks at the `Vary` header when matching, but it looks like we don't set it. 

> To determine if two Requests match, the browser uses more than just the URL. Two requests are considered different if they have different query strings, Vary headers, or HTTP methods (GET, POST, PUT, etc.). https://web.dev/articles/cache-api-quick-guide#retrieving

SgtPooki:
> The Cache API does not support URLs with the ipfs:// protocol scheme:

you're right.. I forgot service worker cache requires inputs to be a request object

src/sw.ts
```diff
@@ -101,7 +105,46 @@ self.addEventListener('fetch', (event) => {
     // intercept and do our own stuff...
     event.respondWith(fetchHandler({ path: url.pathname, request }))
   } else if (isSubdomainRequest(event)) {
-    event.respondWith(fetchHandler({ path: url.pathname, request }))
+    const isMutable = protocol === 'ipns'
+    const cacheKey = `${event.request.url}-${event.request.headers.get('Accept') ?? ''}`
+    const cacheName = isMutable ? MUTABLE_CACHE : IMMUTABLE_CACHE
+    log('helia-sw: cache name: %s | key: %s', cacheName, cacheKey)
+
+    event.respondWith((async () => {
```
SgtPooki: we should move this into a separate function

src/sw.ts
```diff
@@ -101,7 +105,46 @@ self.addEventListener('fetch', (event) => {
...
+      const cachedResponse = await cache.match(cacheKey)
+
+      if ((cachedResponse != null) && !hasExpired(cachedResponse)) {
+        // If there is an entry in the cache for event.request,
+        // then response will be defined and we can just return it.
+        log('helia-ws: cached response HIT for %s (expires: %s) %o', cacheKey, cachedResponse.headers.get('Expires'), cachedResponse)
+
+        trace('helia-ws: updating cache for %s in the background', cacheKey)
+        // 👇 update cache in the background wihtout awaiting
+        fetchHandler({ path: url.pathname, request })
```
SgtPooki: I think we should move the cache logic into the fetchHandler

src/sw.ts
```diff
@@ -177,6 +220,37 @@ function isSwAssetRequest (event: FetchEvent): boolean {
   return isActualSwAsset
 }
 
+/**
+ * Set the expires header with a timestamp base
+ */
+function setExpiryHeader (response: Response, ttlSeconds: number = 3600): Response {
```
SgtPooki: `setExpiryHeader` modifies respToCache but also returns a `Response`. We should change response type to `void` since we're not using the response

SgtPooki (commented):
looks mostly good, we need to figure out https://github.com/ipfs/helia-verified-fetch/issues/17 tho

2color:
> we need to figure out https://github.com/ipfs/helia-verified-fetch/issues/17 tho

Yeah. 

That can be done in parallel, i.e. no need to block this PR, since the [Cache API takes precedence over the browser HTTP cache.](https://web.dev/articles/service-worker-caching-and-http-caching#overview_of_caching_flow).

src/sw.ts
2color: Moved it into a separate function.

src/sw.ts
2color: Have a look at how I structured it now, but I think it makes sense to keep the two functions separate, where fetchHandler actually handles fetching with helia while getResponseFromCacheOrFetch handles caching and delegating to fetchHandler.

2color:
Pretty obvious but worth noting here that depending on whether the origin has `ipns` or `ipfs`, every resource handled by the service worker will either be mutable or immutable. 

For example, when opening `vitalik-eth.ipns.inbrowser.link` all of the resources requested as a path on the origin are considered mutable.

On the other hand, all resources `bafybeiawq7pbt4krnopfmcvymvp2uz4ohibd5p7ugskkybvdmwa2v7evpy.ipfs.inbrowser.link` are **immutable**.

src/sw.ts
```diff
@@ -177,13 +186,93 @@ function isSwAssetRequest (event: FetchEvent): boolean {
   return isActualSwAsset
 }
 
+/**
+ * Set the expires header on a response object to a timestamp based on the passed ttl interval
+ */
+function setExpiresHeader (response: Response, ttlSeconds: number = 3600): void {
+  const expirationTime = new Date(Date.now() + ttlSeconds * 1000)
+
+  response.headers.set('Expires', expirationTime.toUTCString())
```
2color: It may be sensible to make this cache header explicitly SW specific (originally suggested in https://phyks.me/2019/01/manage-expiration-of-cached-assets-with-service-worker-caching.html)
```suggestion
  response.headers.set('sw-cache-expires', expirationTime.toUTCString())
```

SgtPooki:
> For example, when opening vitalik-eth.ipns.inbrowser.link all of the resources requested as a path on the origin are considered mutable.

Yea, it would be nice if we could map the ipns name to a CID and re-route requests to ipfs://cid so that they're cached.

src/sw.ts
```diff
@@ -98,10 +108,9 @@ self.addEventListener('fetch', (event) => {
   }
 
   if (isRootRequestForContent(event)) {
-    // intercept and do our own stuff...
     event.respondWith(fetchHandler({ path: url.pathname, request }))
   } else if (isSubdomainRequest(event)) {
-    event.respondWith(fetchHandler({ path: url.pathname, request }))
+    event.respondWith(getResponseFromCacheOrFetch(event))
```
SgtPooki, line 113: :+1:

src/sw.ts
```diff
@@ -177,13 +186,93 @@ function isSwAssetRequest (event: FetchEvent): boolean {
...
+
+  response.headers.set('Expires', expirationTime.toUTCString())
+}
+
+/**
+ * Checks whether a cached response object has expired by looking at the expires header
+ * Note that this ignores the Cache-Control header since the expires header is set by us
+ */
+function hasExpired (response: Response): boolean {
+  const expiresHeader = response.headers.get('Expires')
```
```suggestion
  const expiresHeader = response.headers.get('sw-cache-expires')
```

src/sw.ts
SgtPooki: that makes sense

src/sw.ts
```diff
@@ -177,13 +186,93 @@ function isSwAssetRequest (event: FetchEvent): boolean {
...
+  }
+
+  const expires = new Date(expiresHeader)
+  const now = new Date()
+
+  if (expires < now) {
+    return true
+  }
+
+  return false
```
```suggestion
  const now = new Date()
  return expires < now
```

src/sw.ts
```diff
@@ -197,6 +286,13 @@ async function fetchHandler ({ path, request }: FetchHandlerArg): Promise<Respon
     })
   }
 
+  /**
+   * > Any global variables you set will be lost if the service worker shuts down.
+   *
+   * @see https://developer.chrome.com/docs/extensions/develop/concepts/service-workers/lifecycle
+   */
+  verifiedFetch = verifiedFetch ?? await getVerifiedFetch()
+
```
SgtPooki, line 301: :+1: I was thinking about doing this the other day

src/sw.ts
```diff
@@ -177,13 +186,93 @@ function isSwAssetRequest (event: FetchEvent): boolean {
...
+  }
+
+  // Clone the response since streams can only be consumed once.
+  const respToCache = response.clone()
+
+  if (isMutable) {
+    trace('helia-ws: setting expires header on response key %s before storing in cache', cacheKey)
+    // 👇 Set expires header to an hour from now for mutable (ipns://) resources
+    // Note that this technically breaks HTTP semantics, whereby the cache-control max-age takes precendence
+    // Seting this header is only used by the service worker asd a mechanism similar to stale-while-revalidate
```
```suggestion
    // Setting this header is only used by the service worker using a mechanism similar to stale-while-revalidate
```

src/sw.ts
```diff
@@ -177,13 +186,93 @@ function isSwAssetRequest (event: FetchEvent): boolean {
...
+    return
+  }
+
+  // Clone the response since streams can only be consumed once.
+  const respToCache = response.clone()
+
+  if (isMutable) {
+    trace('helia-ws: setting expires header on response key %s before storing in cache', cacheKey)
+    // 👇 Set expires header to an hour from now for mutable (ipns://) resources
+    // Note that this technically breaks HTTP semantics, whereby the cache-control max-age takes precendence
```
SgtPooki, line 272: FYI service worker cache takes precedence: https://web.dev/articles/service-worker-caching-and-http-caching

but good to call out that it may differ from `cache-control`

src/sw.ts
```diff
@@ -177,13 +186,93 @@ function isSwAssetRequest (event: FetchEvent): boolean {
...
+    // Seting this header is only used by the service worker asd a mechanism similar to stale-while-revalidate
+    setExpiresHeader(respToCache, 3600)
+  }
+
+  log('helia-ws: storing cache key %s in cache', cacheKey)
+  void cache.put(cacheKey, respToCache)
```
```suggestion
  await cache.put(cacheKey, respToCache)
```

control flow should be done in higher level fetchHandler or getResponseFromCacheOrFetch

src/sw.ts
```diff
@@ -177,13 +186,93 @@ function isSwAssetRequest (event: FetchEvent): boolean {
...
+  const url = new URL(event.request.url)
+  const isMutable = protocol === 'ipns'
+  const cacheKey = `${event.request.url}-${event.request.headers.get('Accept') ?? ''}`
+  trace('helia-sw: cache key: %s', cacheKey)
+  const cache = await caches.open(isMutable ? MUTABLE_CACHE : IMMUTABLE_CACHE)
+  const cachedResponse = await cache.match(cacheKey)
+
+  if ((cachedResponse != null) && !hasExpired(cachedResponse)) {
+  // If there is an entry in the cache for event.request,
+  // then response will be defined and we can just return it.
```
SgtPooki: spacing here seems off
```suggestion
    // If there is an entry in the cache for event.request,
    // then response will be defined and we can just return it.
```

src/sw.ts
```diff
@@ -177,13 +186,93 @@ function isSwAssetRequest (event: FetchEvent): boolean {
...
+  }
+
+  // 👇 fetch because no cached response was found
+  const response = await fetchHandler({ path: url.pathname, request: event.request })
+
+  void storeReponseInCache({ response, isMutable, cache, cacheKey }).catch(err => {
+    err('helia-ws: failed storing response in cache for %s', cacheKey, err)
+  })
+
+  return response
```
SgtPooki: I feel like we could combine these so we don't have two potential places where cache storing happens and could fail

SgtPooki (commented):
leaving some comments.. im going to check the code out locally and play with it

src/sw.ts
```diff
@@ -177,13 +186,93 @@ function isSwAssetRequest (event: FetchEvent): boolean {
...
+  // then response will be defined and we can just return it.
+    log('helia-ws: cached response HIT for %s (expires: %s) %o', cacheKey, cachedResponse.headers.get('Expires'), cachedResponse)
+
+    trace('helia-ws: updating cache for %s in the background', cacheKey)
+    // 👇 update cache in the background (don't await)
+    fetchHandler({ path: url.pathname, request: event.request })
+      .then(async response => storeReponseInCache({ response, isMutable, cache, cacheKey }))
+      .catch(err => {
+        err('helia-ws: failed updating response in cache for %s in the background', cacheKey, err)
+      })
```
SgtPooki: do we need to call `fetchHandler` and update the cache if we already have a cached response?

src/sw.ts
```diff
@@ -177,13 +186,93 @@ function isSwAssetRequest (event: FetchEvent): boolean {
...
+  // If there is an entry in the cache for event.request,
+  // then response will be defined and we can just return it.
+    log('helia-ws: cached response HIT for %s (expires: %s) %o', cacheKey, cachedResponse.headers.get('Expires'), cachedResponse)
+
+    trace('helia-ws: updating cache for %s in the background', cacheKey)
+    // 👇 update cache in the background (don't await)
+    fetchHandler({ path: url.pathname, request: event.request })
+      .then(async response => storeReponseInCache({ response, isMutable, cache, cacheKey }))
+      .catch(err => {
+        err('helia-ws: failed updating response in cache for %s in the background', cacheKey, err)
```
```suggestion
        error('helia-ws: failed updating response in cache for %s in the background', cacheKey, err)
```

src/sw.ts
SgtPooki: yep, totally what I was thinking :)

src/sw.ts
SgtPooki: looking at the code more locally, we don't need to fetch from the network at all on cache-hit.  Since we're currently fighting rate limits from the trustless gateway, we could just not do network requests. The likelihood of something updating on IPFS is very low, whereas stale-while-revalidate is good for frequently updated content that you want a quick response for.

src/sw.ts
SgtPooki: resolved in #109

src/sw.ts
SgtPooki: resolved in #109. stale-while-revalidate only for ipns cache HIT

src/sw.ts
SgtPooki: resolved in #109

src/sw.ts
SgtPooki: resolved in #109

src/sw.ts
SgtPooki: resolved in #109

src/sw.ts
SgtPooki: resolved in #109

src/sw.ts
SgtPooki: resolved in #109

src/sw.ts
SgtPooki: resolved in #109

SgtPooki (approved):

src/sw.ts
lidel: Long term, if we can avoid inventing things specific to service worker, and reuse HTTP semantics, that is better for verified-fetch being useful in more contexts, and easier to test (if these are set upstream, places like https://github.com/ipfs/helia-http-gateway get these headers for free, and we also increase surface of things that can be tested there).

My initial thought here was that if we reuse semantically meaningful `Expires` and set it only if not set yet, in the future it can be set inside `verified-fetch`, and not only we don't need to invent special header here, but also improve the upstream library and the code here will be smaller.  

But fine to go with custom header here for now.

2color: @lidel Because the SW Cache API always takes precedence over browser HTTP cache and and `verified-fetch` [will be returning the `Cache-Control` header](https://github.com/ipfs/helia-verified-fetch/pull/19), the `Expires` header would be: no-op. 

Even if we were to use the `Expires` header, by respecting it when there's a `Cache-Control` header, we'd be breaking from HTTP semantics. This is why I think it makes sense to make it clear that this header is purely for SW purposes.

SgtPooki (approved):

PR #94, opened 2024-03-08
fix: config loading on subdomains waits for updated config

SgtPooki:
## Title

fix: config loading on subdomains waits for updated config

## Description

Minor updates to https://github.com/ipfs-shipyard/helia-service-worker-gateway/pull/89

Addresses the issue brought up in #89:

> Config passing between origins needs to be more carefully tested. It seems like in some cases this is not working properly, e.g. when auto-reload is set up and there isn't enough time for the message to be passed from the root to the subdomain origin.

## Notes & open questions

- uses event.waitUntil to ensure that verifiedFetch is set properly before executing requests
- uses commsChannel.messageAndWaitForResponse to ensure that SW is ready to handle requests before executing them

### Demo

https://github.com/ipfs-shipyard/helia-service-worker-gateway/assets/1173416/7b55d3b2-4090-4285-ab5f-fa6a69fb84a7

## Change checklist


SgtPooki (commented):
self review

SgtPooki (commented):
self review

SgtPooki (commented):
self review

src/lib/config-db.ts
```diff
@@ -71,8 +71,9 @@ export async function loadConfigFromLocalStorage (): Promise<void> {
 }
 
 export async function setConfig (config: ConfigDb): Promise<void> {
-  log('config-debug: setting config', config)
-  debugLib.enable(config.debug ?? '')
+  debugLib.enable(config.debug ?? '') // set debug level first.
+  log('config-debug: setting config %O for domain %s', config, window.location.origin)
```
SgtPooki, line 75: set debug level before calling log method so debug level is taken into consideration immediately.

src/redirectPage.tsx
```diff
@@ -27,20 +27,13 @@ export default function RedirectPage (): JSX.Element {
     async function doWork (config: ConfigDb): Promise<void> {
       try {
         await setConfig(config)
-        // TODO: use channel.messageAndWaitForResponse to ensure that the config is loaded before proceeding.
-        channel.postMessage({ target: 'SW', action: 'RELOAD_CONFIG' })
+        // TODO: show spinner / disable buttons while waiting for response
```
SgtPooki, line 30: things happen pretty fast so i don't think we want a spinner or other things, but maybe?

SgtPooki (commented):
self review of some more minor improvements and manual config update verification

2color:
Great stuff!

PR #96, opened 2024-03-08
feat: allow loading config page from subdomain after sw registration

SgtPooki:
## Title

feat: pass through helia service worker assets & config page requests

## Description

This fix allows you to reload the redirectPage (config in iframe) on any service worker registered in a subdomain by simply appending `#/ipfs-sw-config` to the URL.

It also passes through requests of the service worker source javascript to the server, which it didn't previously do.

**NOTE:** subdomains don't automatically refresh the service worker.. so we need some clever logic for checking for service worker assets, and then redirecting to /#/ipfs-sw-config again. We can handle this later? I will call this out in https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/31

## Notes & open questions

- [x] Need to double check how this works on servers hosted on ipfs. Do we need some _redirects file magic for assets? I think it will work as is (because the assets DO exist on the server & in the content folder..), but need to test...
    - We do not need to do any special redirects handling. since the assets exist, and SW is passing them through, they will be served.
- [x] We can probably remove `/config /#/ipfs-sw-config 302` from the _redirects file now?

This is built on top of https://github.com/ipfs-shipyard/helia-service-worker-gateway/pull/95 which is built on top of https://github.com/ipfs-shipyard/helia-service-worker-gateway/pull/89

### Demo

https://github.com/ipfs-shipyard/helia-service-worker-gateway/assets/1173416/98d387eb-d0e2-4a2e-aebc-4c24c14d1f78

## Change checklist


SgtPooki:
Note: this changes the service worker name, so clearing existing service worker is required with the current code. Details: https://web.dev/learn/pwa/service-workers#updating_a_service_worker

That would result in the following error:
```
Failed to update a ServiceWorker for scope ('http://localhost:3000/') with script ('http://localhost:3000/src_ipfs-sw-_ts.bundle.js'): A bad HTTP response code (404) was received when fetching the script.
TypeError: Failed to update a ServiceWorker for scope ('http://localhost:3000/') with script ('http://localhost:3000/src_ipfs-sw-_ts.bundle.js'): A bad HTTP response code (404) was received when fetching the script.
```

Some ideas: 

- create a fake sw.js in the dist folder to unregister it if it exists & remove later (how do we know when no one else has it registered?)
- update service worker registration code to check for old sw registration (is this possible) and unregister it.
- add code to automatically unregister any registered service workers, then re-add the new url (this could cause implications for ipfs sites service workers)
- dont change the sw name (webpack config fu?) and spy on sw asset with explicit "sw.js" check (this could cause implications for ipfs sites with sw.js files)
- Dont merge this until we have the old-name sw.js updated with code that will handle this for us (how? How do we know when users arent using it anymore)

SgtPooki:
We can create an easy reset for users by adding a #ipfs-sw-reset pathname flag that triggers the following code:

```
self.registration.unregister()
    .then(function() {
      return self.clients.matchAll();
    })
    .then(function(clients) {
      clients.forEach(client => client.navigate(client.url))
    });
```

It would be good to make sure this is in-place before changing the sw name.

public/_redirects
```diff
@@ -1,2 +1,2 @@
-/config /#/config 302
+/config /#/ipfs-sw-config 302
```
2color: How can it redirect to a hash fragment if hash fragments aren't technically sent a server?

SgtPooki: the redirect happens only if it hits the ipfs hosted site backend.

SgtPooki:
screenshot showing loading a subdomain, after adding a tiny little "log('test')" to `sw.ts`, when serving the assets with `npm start`:

image

See the last two lines:

Reloading the page resulted in the browser requesting the new service worker source, and updating it.

public/_redirects
SgtPooki: With that said though, we probably don't need redirect from /config to anything anymore since we're not doing any requests to /config anymore.

`helia-sw` is still needed as a catch all

PR #98, opened 2024-03-11
feat: use separate logger for service worker

2color:
## Description

This PR creates a separate logger instance for the service worker to make it a bit easier for debugging and avoid needing to prefix the log messages in the service worker. 

## Notes & open questions

-

## Change checklist

### DIFF
--- a/src/lib/sw-logger.ts
+++ b/src/lib/sw-logger.ts
@@ -1,6 +1,6 @@
 import { logger } from '@libp2p/logger'
 
-const logObj = logger('service-worker-gateway:ui')
+const logObj = logger('service-worker-gateway:sw')
 
 export const log = logObj
 export const error = logObj.error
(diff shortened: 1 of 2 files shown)


SgtPooki (commented):
we have some functions that are used by the UI and the sw that both use the logger, so we will likely need to pass through a logger option to those methods so the right log prefix is used

2color:
Closed in favour of a different approach discussed in-person where we:
- Use the defaultLogger and call forComponent based on where it's used https://github.com/libp2p/js-libp2p/blob/main/packages/logger/src/index.ts#L171
- For modules that use the logger and are shared by both the SW and FE, we use dep injection to make sure they use a different logger (so as to make it easier to discern where the logs are coming from)

See https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/139

PR #100, opened 2024-03-12
feat: allow explicit sw de-registering + timebomb

### LINKED ISSUE #26: Update mechanisms
## Need

At some point we may want to not reload worker on every page load(?).

We may have a bug which breaks all websites, there may be an important update, or user wants to ensure they have the latest version.

We need reliable mechanisms for forcing reload of SW code.

## Implementation ideas

Below are my initial ideas, feedback welcome. 
I think at minimum, at least short term, we need  (A) and (C).

Global: 

- (A) lo-fi time-bomb: expire worker after some time, force-reload every 24h (at least while we have intensive dev)
- (B) DNSLink beacon checked on each load
  - we could inspect `/ipns/inbrowser.link/sw-version` on sw init, and if set to timestamp/version higher than one in sw code, we force-reload sw to new version.

Per user:

- (C) a `?sw-update` param, allows end user to manually force reload
- (D) we could use `ServiceWorkerRegistration.onupdatefound` it on root domain to softly inform user new version is available, and button they could click to update + have cookie that then informs all subdomains to update as well.

SgtPooki:
I will be deploying these changes to sw.sgtpooki.com so it can be used for testing


SgtPooki:
FYI that the latest changes, i.e. [`fe32830` (#100)](https://github.com/ipfs-shipyard/helia-service-worker-gateway/pull/100/commits/fe32830df87159f089c52eb5ce9c46bce052f0a7), except that the timebomb is set to 30seconds (see below) have been deployed to https://sw.sgtpooki.com

```
╰─ ✔ ❯ git add -p                                                                                                                                     3.73   24.4G   52%   100%  ─╯
Alias tip: g add -p
diff --git a/src/sw.ts b/src/sw.ts
index f859e19..8ec8a02 100644
--- a/src/sw.ts
+++ b/src/sw.ts
@@ -291,7 +291,7 @@ async function isTimebombExpired (): Promise<boolean> {
   firstInstallTime = firstInstallTime ?? await getInstallTimestamp()
   const now = Date.now()
   // max life (for now) is 24 hours
-  const timebomb = 24 * 60 * 60 * 1000
+  const timebomb = 30 * 1000
   return now - firstInstallTime > timebomb
 }
```

This should help with testing. See a demo below:

https://github.com/ipfs-shipyard/helia-service-worker-gateway/assets/1173416/ff29365f-81c4-459a-9681-ea07a6e0efa1

SgtPooki (commented):
self review & local testing of cache && 10sec timebomb

PR #105, opened 2024-03-12
chore(ci): additional pin to web3.storage

lidel:
This PR adds web3.storage pinning by is porting setup from ipfs/ipfs-webui.

Secrets are placed in dedicated `web3.storage` environment which is only exposed to `pin-to-w3` job.
This allows us to add more pinning services and pin to them in parallel (via CAR or CID).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -168,7 +168,7 @@ jobs:
     if: github.ref == 'refs/heads/main'
     needs: build
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


lidel:
I've run it on this branch  (temporary safelist of  this branch) and  works as expected: https://github.com/ipfs-shipyard/helia-service-worker-gateway/actions/runs/8254712298/job/22579548366#step:5:28

Merging to see how smoke-tests behave when DAG is not precached on our HTTP infra.

2color:
Noice!

PR #109, opened 2024-03-14
feat: some updates to #92

SgtPooki:
minor code cleanup and improvements for #92

- chore: PR comments and minor improvements
- fix: use sw-cache-expires
- chore: cleanup hasExpired return
- chore: log on cache miss
- chore: wait for response so we can cache it
- chore: minor cleanup
- chore: minor cleanup
- fix: stale-while-revalidate for ipns

notable changes:

1. stale-while-revalidate only for IPNS content (if people want no network requests, query ipfs resource)
2. sw-cache-expires as called out by @2color 
3. consolidation of fetch + update cache in `fetchAndUpdateCache` function


2color (approved):
Very nice! Thanks for the polish

PR #112, opened 2024-03-14
fix: css styles applied & config page collapse

### LINKED ISSUE #111: bug: css imports are not included in final assets
I used git bisect to discover that https://github.com/ipfs-shipyard/helia-service-worker-gateway/commit/0c81f2aa05db6dbdd64aaaef278fdb11b5c0843a introduced a bug where css imports are not being included.

you can prove this by searching for `background-color: grey` with `<CMD+OPTION+F>` in browser dev tools when running the code locally:

i.e. https://github.com/ipfs-shipyard/helia-service-worker-gateway/blob/220e9cb58c6cae00c0ae50e46f912df68ace60c1/src/components/local-storage-toggle.css#L17

app.css is also not included.

Adding the contents of `local-storage-toggle.css` to the inspector stylesheet in chrome devtools causes the component to show:

image

SgtPooki:
## Title

Fix css styles not being included in final assets & better config page handling

## Description

1. Fix css styles not being included in final assets
2. Remove annoying issue where config page from main domain was always collapsed. It is now only collapsed by default when viewed from a subdomain.
    - ~If a user explicitly wants to see the config from the subdomain, it should not be collapsed by default. Will update shortly.~ This will be done in a separate PR. see https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/113

fixes #111

## Change checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,5 +67,7 @@
     "webpack-dev-server": "^4.15.1",
     "webpack-merge": "^5.8.0"
   },
-  "sideEffects": false
+  "sideEffects": [
+    "*.css"
+  ]
 }
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "css-loader": "^6.10.0",
     "eslint-config-standard-with-typescript": "^34.0.1",
     "html-webpack-plugin": "^5.5.0",
+    "mini-css-extract-plugin": "^2.8.1",
     "npm-run-all": "^4.1.5",
     "patch-package": "^6.5.1",
     "rimraf": "^4.4.1",
--- a/test/node.js
+++ b/test/node.js
@@ -1,6 +1,6 @@
 /* eslint-env mocha */
 import { constants } from 'node:fs'
-import { access } from 'node:fs/promises'
+import { access, readFile } from 'node:fs/promises'
 import { resolve, dirname } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { expect } from 'aegir/chai'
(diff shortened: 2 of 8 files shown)


test/node.js
```diff
@@ -14,4 +14,11 @@ describe('verify-dist', async () => {
    */
     await expect(access(resolve(cwd, '../dist/ipfs-sw-sw.js'), constants.F_OK)).to.not.be.rejected()
   })
+
+  it('has css file with expected content', async () => {
+    await expect(access(resolve(cwd, '../dist/ipfs-sw-styles.css'), constants.F_OK)).to.not.be.rejected()
+    const contents = await readFile(resolve(cwd, '../dist/ipfs-sw-styles.css'), 'utf8')
+    expect(contents).to.include('47vh')
+    expect(contents).to.include('.local-storage-toggle input.status')
+  })
```
SgtPooki, line 23: we will likely want to remove this in the future but i'm going to leave it for now

webpack.config.js
```diff
@@ -237,8 +250,15 @@ const common = {
 
       // Fonts and SVGs: Inline files
       { test: /\.(woff(2)?|eot|ttf|otf|svg|)$/, type: 'asset/inline' },
-
-      { test: /\.(css)$/, use: ['style-loader', 'css-loader'] }
+      {
+        // If you enable `experiments.css` or `experiments.futureDefaults`, please uncomment line below
+        // type: 'javascript/auto',
```
SgtPooki:
```suggestion
```

package.json
```diff
@@ -66,5 +67,7 @@
     "webpack-dev-server": "^4.15.1",
     "webpack-merge": "^5.8.0"
   },
-  "sideEffects": false
+  "sideEffects": [
+    "*.css"
+  ]
```
SgtPooki, line 72: needed this to tell webpack about our css (tree splitting was ignoring css files because we told it we had no side effects..).

IIUC this is a webpack specific thing. see https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

SgtPooki (commented):
self review

PR #138, opened 2024-03-21
feat: quick TTFB & TTI for large content

### LINKED ISSUE #131: bug: loading video in browser native video player doesn't send response quick enough
see https://github.com/ipfs-shipyard/helia-service-worker-gateway/pull/122#issuecomment-2005453039

Basically:

1. when fetching content for a large video, `bafybeidsp6fva53dexzjycntiucts57ftecajcn5omzfgjx57pqfy3kwbq`, the browser doesn't seem to get the response until the full stream is resolved
2. In order to effectively return video files, we need to make sure the response headers are written *immediately* and that the stream resolution is non-blocking.

You can see some examples of how browsers will cancel the first request when it knows it's video content, and then send a byte request immediately after, in this gist: https://gist.github.com/SgtPooki/9efc039d1d349d112d0f505b2ed275c8

SgtPooki:
## Title

feat: quick TTFB & TTI for large content

## Description

This updates helia-service-worker-gateway to @helia/verified-fetch@1.3.0 and ensures we're respecting the request event's AbortSignal.

Summary of changes (thanks @2color):

1. Handle aborts from original fetch event
1. Don't await Cache API puts
1. Respect cache-control: no-cache & pragma: no-cache

## Notes & open questions

after getting https://github.com/ipfs-shipyard/helia-service-worker-gateway/pull/122 and https://github.com/ipfs/helia-verified-fetch/pull/26 merged.. then updating @helia/verified-fetch in service-worker-gateway locally…. large video loading feels much more responsive

For the below, I hardcoded my local kubo node to ensure trustless-gateway.link was not being used.

### big-buck-bunny TTFB/TTI for the code in this PR from kubo node with content:

Brave: consistently < 800ms 
FireFox: 159ms

1. A local kubo gateway configured (that has the content) `npx kubo daemon`
2. cache disabled in the browser.

You can use the below process and code snippet to get TTFB:

1. clear out service worker
3. reload the page (you will see redirect page, make sure config is set up properly)
4. click "load content" or reload.
5. When the video renders, pause it.
6. Run the below snippet in the dev console

```
new PerformanceObserver((entryList) => {
  const videoLoadEntries = performance.getEntriesByType('navigation')
    .filter((entry) => entry.name.includes('bafybeidsp6fva53dexzjycntiucts57ftecajcn5omzfgjx57pqfy3kwbq.ipfs.helia-sw-gateway.localhost'))

  for (const videoLoadEntry of videoLoadEntries) {
    console.log(`TTFB: ${videoLoadEntry.duration}ms for ${videoLoadEntry.name}`)
  }
}).observe({
  type: 'navigation',
  buffered: true
});
```

### big-buck-bunny TTFB/TTI for code in this PR pulling from *fresh* kubo node:

Brave: 4-12s.

see https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/131#issuecomment-2014184623

### big-buck-bunny TTFB/TTI for `main` branch code on kubo node with content

In Brave: Consistently over 1000ms.. with a fresh service worker

In Firefox: idk why but it was returning within a few ms.. old cache/sw that wouldn't go away? https://support.mozilla.org/en-US/kb/clear-cookies-and-site-data-firefox didn't always take

### big-buck-bunny TTFB/TTI for code in `main` from *fresh* kubo node:

Brave: averaging ~70-90 seconds

## Change checklist

### DIFF
--- a/src/sw.ts
+++ b/src/sw.ts
@@ -362,11 +362,11 @@ function shouldCacheResponse ({ event, response }: { event: FetchEvent, response
   }
   const invalidOkResponseCodesForCache = [206]
   if (invalidOkResponseCodesForCache.some(code => code === response.status)) {
-    log('helia-sw-cache-debug: not caching response with status %s', response.status)
+    log('helia-sw: not caching response with status %s', response.status)
     return false
   }
   if (event.request.headers.get('pragma') === 'no-cache' || event.request.headers.get('cache-control') === 'no-cache') {
-    log('helia-sw-cache-debug: request indicated no-cache, not caching')
+    log('helia-sw: request indicated no-cache, not caching')
     return false
   }
 
--- a/src/sw.ts
+++ b/src/sw.ts
@@ -82,6 +82,7 @@ const CURRENT_CACHES = Object.freeze({
 })
 let verifiedFetch: VerifiedFetch
 const channel = new HeliaServiceWorkerCommsChannel('SW')
+const timeoutAbortEventType = 'verified-fetch-timeout'
 const ONE_HOUR_IN_SECONDS = 3600
 const urlInterceptRegex = [new RegExp(`${self.location.origin}/ip(n|f)s/`)]
 const updateVerifiedFetch = async (): Promise<void> => {
@@ -427,7 +428,7 @@ async function fetchHandler ({ path, request, event }: FetchHandlerArg): Promise
   const signal = abortController.signal
   const abortFn = (event: Pick<AbortSignalEventMap['abort'], 'type'>): void => {
     clearTimeout(signalAbortTimeout)
-    if (event?.type !== 'verified-fetch-timeout') {
+    if (event?.type !== timeoutAbortEventType) {
       log.trace('helia-sw: request signal aborted')
       abortController.abort('request signal aborted')
     } else {
@@ -441,7 +442,7 @@ async function fetchHandler ({ path, request, event }: FetchHandlerArg): Promise
    * @todo reduce to 2 minutes?
    */
   const signalAbortTimeout = setTimeout(() => {
-    abortFn({ type: 'verified-fetch-timeout' })
+    abortFn({ type: timeoutAbortEventType })
   }, 5 * 60 * 1000)
   // if the fetch event is aborted, we need to abort the signal we give to @helia/verified-fetch
   event.request.signal.addEventListener('abort', abortFn)
--- a/src/sw.ts
+++ b/src/sw.ts
@@ -428,12 +428,12 @@ async function fetchHandler ({ path, request, event }: FetchHandlerArg): Promise
   const signal = abortController.signal
   const abortFn = (event: Pick<AbortSignalEventMap['abort'], 'type'>): void => {
     clearTimeout(signalAbortTimeout)
-    if (event?.type !== timeoutAbortEventType) {
-      log.trace('helia-sw: request signal aborted')
-      abortController.abort('request signal aborted')
-    } else {
+    if (event?.type === timeoutAbortEventType) {
       log.trace('helia-sw: timeout waiting for response from @helia/verified-fetch')
       abortController.abort('timeout')
+    } else {
+      log.trace('helia-sw: request signal aborted')
+      abortController.abort('request signal aborted')
     }
   }
   /**
--- a/src/sw.ts
+++ b/src/sw.ts
@@ -441,7 +441,7 @@ async function fetchHandler ({ path, request, event }: FetchHandlerArg): Promise
       log.trace('fetchHandler: request headers: %s: %s', key, value)
     })
 
-    const response = await verifiedFetch(verifiedFetchUrl, {
+    const response = verifiedFetch(verifiedFetchUrl, {
       signal,
       headers,
       // TODO redirect: 'manual', // enable when http urls are supported by verified-fetch: https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/62#issuecomment-1977661456
@@ -459,7 +459,7 @@ async function fetchHandler ({ path, request, event }: FetchHandlerArg): Promise
      * the response object, regardless of it's inner content
      */
     clearTimeout(signalAbortTimeout)
-    return response
+    return await response
   } catch (err: unknown) {
     const errorMessages: string[] = []
     if (isAggregateError(err)) {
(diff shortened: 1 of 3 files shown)


SgtPooki:
You can use the below process and code snippet to get TTFB:

1. clear out service worker
2. reload the page (you will see redirect page, make sure config is set up properly)
3. click "load content" or reload.
4. When the video renders, pause it.
5. Run the below snippet in the dev console

```
const videoLoadEntry = performance.getEntriesByType('resource').filter((entry) => entry.name.includes('bafybeidsp6fva53dexzjycntiucts57ftecajcn5omzfgjx57pqfy3kwbq.ipfs.helia-sw-gateway.localhost'))[0] /** get the very first entry, because there may be multiple if you don't stop the video quick enough */
console.log(`TTFB: ${videoLoadEntry.responseStart - videoLoadEntry.requestStart}ms`)
```

SgtPooki (commented):
self review

SgtPooki:
changes deployed on https://bafybeidsp6fva53dexzjycntiucts57ftecajcn5omzfgjx57pqfy3kwbq.ipfs.sw.sgtpooki.com/

use https://support.mozilla.org/en-US/kb/clear-cookies-and-site-data-firefox to clear data on firefox.

TTFB on firefox from sw.sgtpooki.com is ~14000ms

src/sw.ts
```diff
@@ -356,10 +356,25 @@ async function getResponseFromCacheOrFetch (event: FetchEvent): Promise<Response
...
 
-const invalidOkResponseCodesForCache = [206]
-async function storeReponseInCache ({ response, isMutable, cacheKey }: StoreReponseInCacheOptions): Promise<void> {
-  // 👇 only cache successful responses
-  if (!response.ok || invalidOkResponseCodesForCache.some(code => code === response.status)) {
+function shouldCacheResponse ({ event, response }: { event: FetchEvent, response: Response }): boolean {
+  if (!response.ok) {
+    return false
+  }
+  const invalidOkResponseCodesForCache = [206]
```
2color:
```suggestion
  const statusCodesToNotCache = [206]
```
This name is a bit confusing. Using an array for this seems like a premature optimisation here given we only have one code.

src/sw.ts
```diff
@@ -356,10 +356,25 @@ async function getResponseFromCacheOrFetch (event: FetchEvent): Promise<Response
...
+  if (invalidOkResponseCodesForCache.some(code => code === response.status)) {
+    log('helia-sw: not caching response with status %s', response.status)
+    return false
+  }
+  if (event.request.headers.get('pragma') === 'no-cache' || event.request.headers.get('cache-control') === 'no-cache') {
```
2color, line 369: Nice!

src/sw.ts
```diff
@@ -407,8 +423,29 @@ async function fetchHandler ({ path, request }: FetchHandlerArg): Promise<Respon
    * * https://bugs.chromium.org/p/chromium/issues/detail?id=823697
    * * https://bugzilla.mozilla.org/show_bug.cgi?id=1394102
    */
-  // 5 minute timeout
-  const signal = AbortSignal.timeout(5 * 60 * 1000)
+  const abortController = new AbortController()
+  const signal = abortController.signal
+  const abortFn = (event: Pick<AbortSignalEventMap['abort'], 'type'>): void => {
+    clearTimeout(signalAbortTimeout)
```
2color, line 430: I think you'll need to move `signalAbortTimeout` above this function, since `const` statements are not hoisted like `vars` and might not be available in the closure when you define `abortFn`

src/sw.ts
```diff
@@ -407,8 +423,29 @@ async function fetchHandler ({ path, request }: FetchHandlerArg): Promise<Respon
...
+    if (event?.type !== 'verified-fetch-timeout') {
```
2color: For readability: maybe switch the conditions and check for the positive case first

Also maybe move `verified-fetch-timeout` to a constant

2color (approved):
Barring a bunch of nits this is great! 

Summary of changes (for me):
1. Handle aborts from original fetch event
2. Don't await Cache API puts (TIL)
3. Respect cache-control: no-cache

(Don't cache 206s was already there but new to my eyes)

src/sw.ts
```diff
@@ -407,8 +423,29 @@ async function fetchHandler ({ path, request }: FetchHandlerArg): Promise<Respon
...
+  /**
+   * five minute delay to get the initial response.
+   *
+   * @todo reduce to 2 minutes?
+   */
+  const signalAbortTimeout = setTimeout(() => {
+    abortFn({ type: 'verified-fetch-timeout' })
+  }, 5 * 60 * 1000)
+  // if the fetch event is aborted, we need to abort the signal we give to @helia/verified-fetch
+  event.request.signal.addEventListener('abort', abortFn)
```
2color, line 448: We might want to remove this eventListener after the verifiedFetch promise is resolved/rejected to avoid memory leaks.

src/sw.ts
SgtPooki: idk that it's an optimization.. probably more of a slight perf hit due to looping the array instead of checking a single number, but maybe it's optimizing for future dev work

src/sw.ts
SgtPooki: IIUC correctly the `event.request` is dead (and all things attached to it) as soon as the request is consumed, but that's a good CYA improvement

SgtPooki: updated.

src/sw.ts
SgtPooki: updated

src/sw.ts
SgtPooki: Actually, we can't remove the event listener for this in a way that makes sense, because the browser controls the request and we would need to remove the listener after consumption of the response body.

src/sw.ts
SgtPooki: const is "hoisted" into it's scoped closure (i.e. available in the scopes lookup table), which `abortFn` is also a part of

short example:

```
(() => {
    const abortFn = () => { console.log(varDefinedAfterFn) }
    const varDefinedAfterFn = 'hello world'
    abortFn() // hello world
})()
```

If they were in different scopes.. it wouldn't:

```
(() => {
    const abortFn = () => { console.log(varDefinedAfterFn) }
    {const varDefinedAfterFn = 'hello world'}
    abortFn() // Uncaught ReferenceError: varDefinedAfterFn is not defined
})()
```

2color: Thanks for clarifying that.

PR #154, opened 2024-03-25
chore: add permissive LICENSE.md

### LINKED ISSUE #116: Missing LICENSE
We need to 

- [x] Port dual-licensing we do in Kubo:
  - https://github.com/ipfs/kubo/blob/master/README.md?plain=1#L433-L438
  - https://github.com/ipfs/kubo/blob/master/LICENSE
- [x] Get sign-off from everyone that did commits so far

2color:
Fixes #116


SgtPooki:
We should use the same pattern as libp2p with two license files

LICENSE.md
lidel, line 1: @2color any reason why we used different syntax than we have in Kubo and Helia repos?

I was under impression we want to copy&paste:
- https://github.com/ipfs/helia/blob/main/LICENSE (we could add SPDX tags here and in README)
- https://github.com/ipfs/helia/blob/main/LICENSE-MIT
- https://github.com/ipfs/helia/blob/main/LICENSE-APACHE

lidel (changes requested):
@2color thanks! Just a small concern around making this repo use different "syntax" than Kubo and Helia. 

Is the idea to update them? If not, see comment below.

LICENSE.md
aschmahmann: We can use a single one, no https://github.com/ipfs/ipfs-repository-template/blob/c0e7956eae1c7541e11fc24c24058085fc8f3339/LICENSE.md? If not should we change the repo template?

SgtPooki: js-libp2p and helia have been using separate ones for a while, we should stick to that pattern.

2color: I don't have strong opinions. I also don't want to waste time on this.

I've stuck to the same convention as Helia and added the SPDX tag in the `LICENSE` file. 

Hopefully that should keep everyone happy.

SgtPooki (approved):

SgtPooki:
@lidel, what should we do for the dev sign-off process for previous committers?

2color:
I think we can just continue collecting approvals here. 

@lidel @whizzzkid @lidel @achingbrain @MarcoPolo @aschmahmann 

Can you please review this and approve to confirm you are ok with the license update?

lidel (commented):
I accept these license terms.

ps. we should get approvals before merging, but as long everyone replies here we are good and don't have to revert :)

SgtPooki:
I accept these license terms.

aschmahmann:
I accept these license terms.

whizzzkid:
I accept these license terms.

achingbrain:
I accept these license terms.

LICENSE.md
achingbrain: Urgh, what a pain.  aegir-managed projects should be using these files: https://github.com/ipfs/aegir/tree/master/src/check-project/files which I think were taken from go-ipfs in the dim and distant past.  I may be wrong, memory is a funny thing.

SgtPooki:
just need Marco and I think we're good

MarcoPolo:
I accept these license terms.

LICENSE.md
2color: @achingbrain what implication does it have if the license files are different?

For example, in this repo I added the `SPDX-License-Identifier: Apache-2.0 OR MIT` tag to `LICENSE`.

PR #202, opened 2024-04-16
feat: helper-ui input-validator update

### LINKED ISSUE #66: Input format validation doesn't match up with expected format
Hitting https://ipfs.sw.sgtpooki.com tells you to use a `/ipfs/bafk.../path/to/file`-style IPFS path:

image

Filling in the field with an IPFS path tells you to use a `ipfs://`-style IPFS URI:

image

Switching to an IPFS URI tells you to use an IPFS path:

image

SgtPooki:
## Title

feat: helper-ui input-validator update

## Description

Fixes https://github.com/ipfs-shipyard/service-worker-gateway/issues/66

## Notes & open questions

the input validator now accepts the same urls as verified-fetch (more actually since https? is not required) and provides more helpful information

## Change checklist


SgtPooki (commented):
self review

2color (commented):
From a code perspective looks really clean and correct for all the formats accepted. all the tests.

A couple points regarding UX:
-  The help text with all the accepted formats is overwhelming. My opinion is that we should be a little more prescriptive about the preferred/conventional format while gracefully accepting others (without advertising all of the the graceful cases). The formats with `example.com` are particularly confusing. 
- I'm somewhat surprised that we don't accept `ipfs://` and `ipns://` paths. I thought those would be considered the canonical form. I'm also that the `/ipfs/..` and `/ipns/..` are also common conventions in Kubo. 
- Perhaps instead of adding `/path` to each example, you omit it from all and update the note to say that you can suffix any path with a path?
 
Screenshot 2024-04-16 at 3 03 20 pm

SgtPooki:
> The help text with all the accepted formats is overwhelming. My opinion is that we should be a little more prescriptive about the preferred/conventional format while gracefully accepting others (without advertising all of the the graceful cases). The formats with example.com are particularly confusing.

Ok good to know, do you or @lidel have recommended example text we should include here?

> I'm somewhat surprised that we don't accept ipfs:// and ipns:// paths. I thought those would be considered the canonical form. I'm also that the /ipfs/.. and /ipns/.. are also common conventions in Kubo.

😲 I missed that. wow. Will add.

> Perhaps instead of adding /path to each example, you omit it from all and update the note to say that you can suffix any path with a path?

good advice

SgtPooki:
Ok I've got the help text as a table, how is this:

image

lidel:
It still feels overwhelming. No need to flood user with all variants and permutations.
We just want to hint how valid value type looks like + maybe link to docs.

Maybe simplify closer to:

> ⚠️ Invalid address, correct it and try again.
> 
> ℹ️ For reference, accepted formats are:
> - UNIX-like Content Path (`/ipfs/cid/..`, `/ipns/id/..`), 
> - HTTP Gateway URL (`https://ipfs.io/ipfs/cid..`, `https://cid.ipfs.dweb.link/..`), 
> - Native IPFS URL (`ipfs://cid/..`, `ipns://id/..`)
>
> Learn more at [Addressing IPFS on the Web](https://docs.ipfs.tech/how-to/address-ipfs-on-web)

SgtPooki:
> Maybe simplify closer to:

Ok will do

SgtPooki:
that doesn't look much better to me tbh:

image


I think we should have a single example per "Type" we want to show:

image


but there's also some alignment issues with that, so, back to a table but without the header:

image

lidel:
Agree, one example per type looks better :+1: 

Lgtm, but while we are at it, we could improve two things:
- Rename input label from "CID (and path)" to something like "CID, Content Path, or URL" 
  - Since it has "CID", add "CID" type (support pasing standalone CID without `/ipfs/` prefix)

SgtPooki:
> * Rename input label from "CID (and path)" to something like "CID, Content Path, or URL"
>     * Since it has "CID", add "CID" type (support pasing standalone CID without /ipfs/ prefix)

Done since https://github.com/ipfs-shipyard/service-worker-gateway/pull/202/commits/76ae50cdaa2500256beab38d38c2eddb32a6c87c & https://github.com/ipfs-shipyard/service-worker-gateway/pull/202/commits/30e03825b1c23d40e1d8d3a0ed1efec22e046efa

SgtPooki:
merging due to implicit approvals via emojis on comments

PR #215, opened 2024-04-23
fix: ipfs-hosted redirects are not infinite

### LINKED ISSUE #211: bug: infinite redirect
on first-hit:

https://inbrowser.dev/ipns/app.keom.io
https://inbrowser.dev/ipns/vitalik.eth
https://inbrowser.dev/ipfs/bafkqabtimvwgy3yk // "hello" inline CID

SgtPooki:
## Title

fix: ipfs-hosted redirects are not infinite

## Description

Fixes https://github.com/ipfs-shipyard/service-worker-gateway/issues/211

### Summary of changes

1. Adds ipfs-gateway.js script that uses existing reverse-proxy (some updates were needed) to host static site via local kubo
1. Updates `ipfs-hosted` tests in first-hit.test.ts to use the above test server, so we are no longer only getting a ?helia-sw request. This should get us 99% closer to actual functionality when hosted in environments like inbrowser.tld
1. Resolves issues with infinite redirects by updating `src/pages/redirects-interstitial.tsx` and `src/pages/redirect-page.tsx`
1. Moves out service worker & config provider from the initial landing page into each sub-page, which reduces the total assets sent on first-hit for each scenario.
1. updates reverse-proxy logging to use @libp2p/logger, ~~but it's not actually working for some reason~~

## Notes & open questions

1. ~~I could use an extra set of eyes on the updated use of @libp2p/logger in reverse-proxy.js~~ proxy-server logging fixed with [`688594e` (#215)](https://github.com/ipfs-shipyard/service-worker-gateway/pull/215/commits/688594e4824a91fdf206cc7c8d36494254aa5835)
2. Callout that maybe we don't want to display in `RedirectsInterstitial` because it's just going to be quick Flashes that aren't really visible, but on gateways where providing the assets are slow, it will be informative.
3. If you set the playback speed to very slow on the demo video you can see the redirect cycle pretty clearly

### Demo showing fix:

https://github.com/ipfs-shipyard/service-worker-gateway/assets/1173416/d06ebbd5-1270-4a21-9e5f-ba6a8f870791

## Change checklist

### DIFF
--- a/src/lib/translate-ipfs-redirect-url.ts
+++ b/src/lib/translate-ipfs-redirect-url.ts
@@ -1,6 +1,7 @@
 /**
- * If you host helia-service-worker-gateway on an IPFS domain, the redirects file will route some requests from
- * `<domain>/<wildcard-splat>` to `https://<domain>/?helia-sw=<wildcard-splat>`.
+ * If you host helia-service-worker-gateway with an IPFS gateway, the _redirects file will route some requests from
+ * `<domain>/<wildcard-splat>` to `https://<domain>/?helia-sw=<wildcard-splat>` when they hit the server instead of
+ * the service worker. This only occurs when the service worker is not yet registered.
  *
  * This function will check for "?helia-sw=" in the URL and modify the URL so that it works with the rest of our logic
  */
--- a/src/lib/translate-ipfs-redirect-url.ts
+++ b/src/lib/translate-ipfs-redirect-url.ts
@@ -1,0 +1,15 @@
+/**
+ * If you host helia-service-worker-gateway on an IPFS domain, the redirects file will route some requests from
+ * `<domain>/<wildcard-splat>` to `https://<domain>/?helia-sw=<wildcard-splat>`.
+ *
+ * This function will check for "?helia-sw=" in the URL and modify the URL so that it works with the rest of our logic
+ */
+export function translateIpfsRedirectUrl (urlString: string): URL {
+  const url = new URL(urlString)
+  const heliaSw = url.searchParams.get('helia-sw')
+  if (heliaSw != null) {
+    url.searchParams.delete('helia-sw')
+    url.pathname = heliaSw
+  }
+  return url
+}
--- a/src/pages/redirects-interstitial.tsx
+++ b/src/pages/redirects-interstitial.tsx
@@ -1,34 +1,41 @@
-import React from 'preact/compat'
+import React, { useEffect } from 'preact/compat'
+import { findOriginIsolationRedirect } from '../lib/path-or-subdomain.js'
+import { translateIpfsRedirectUrl } from '../lib/translate-ipfs-redirect-url.js'
+import RedirectPage from './redirect-page'
 
 /**
  * This page is only used to capture the ?helia-sw=/ip[fn]s/blah query parameter that
  * is used by IPFS hosted versions of the service-worker-gateway when non-existent paths are requested.
+ * This will only redirect if the URL is for a subdomain
  */
 export default function RedirectsInterstitial (): React.JSX.Element {
-  const windowLocation = translateIpfsRedirectUrl(window.location.href)
-  if (windowLocation.href !== window.location.href) {
-  /**
-   * We're at a domain with ?helia-sw=, we can reload the page so the service worker will
-   * capture the request
-   */
-    window.location.replace(windowLocation.href)
+  const [subdomainRedirectUrl, setSubdomainRedirectUrl] = React.useState<string | null>(null)
+  const [isSubdomainCheckDone, setIsSubdomainCheckDone] = React.useState<boolean>(false)
+  useEffect(() => {
+    async function doWork (): Promise<void> {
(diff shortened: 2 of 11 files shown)


src/lib/translate-ipfs-redirect-url.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * If you host helia-service-worker-gateway on an IPFS domain, the redirects file will route some requests from
```
lidel:
```suggestion
 * If you host helia-service-worker-gateway with IPFS gateway, the _redirects file will route some requests from
```

src/lib/translate-ipfs-redirect-url.ts
```diff
@@ -0,0 +1,15 @@
...
+ * `<domain>/<wildcard-splat>` to `https://<domain>/?helia-sw=<wildcard-splat>`.
```
SgtPooki:
```suggestion
 * If you host helia-service-worker-gateway on an IPFS domain, the _redirects file will route some requests from
 * `<domain>/<wildcard-splat>` to `https://<domain>/?helia-sw=<wildcard-splat>` if they hit the server instead of the service-worker.
```

lidel (commented):
Thanks! Had no bandwidth today to review code, but if this data point is useful: I did run it locally against static http server and `?helia-sw=/ipns/docs.ipfs.tech` and it no longer redirects to `/ipns/docs.ipfs.tech` so the infinite loop from https://github.com/ipfs-shipyard/service-worker-gateway/issues/211#issuecomment-2072428579 does not occur anymore.

test-e2e/ipfs-gateway.js
```diff
@@ -0,0 +1,114 @@
...
+await $(execaOptions)`${kuboBin} config Addresses.Gateway /ip4/127.0.0.1/tcp/${gatewayPort}`
+await $(execaOptions)`${kuboBin} config Addresses.API /ip4/127.0.0.1/tcp/0`
+await $(execaOptions)`${kuboBin} config --json Bootstrap ${JSON.stringify([])}`
+await $(execaOptions)`${kuboBin} config --json Swarm.DisableNatPortMap true`
+await $(execaOptions)`${kuboBin} config --json Discovery.MDNS.Enabled false`
+await $(execaOptions)`${kuboBin} config --json Gateway.NoFetch true`
+await $(execaOptions)`${kuboBin} config --json Gateway.DeserializedResponses true`
+await $(execaOptions)`${kuboBin} config --json Gateway.ExposeRoutingAPI false`
+await $(execaOptions)`${kuboBin} config --json Gateway.HTTPHeaders.Access-Control-Allow-Origin ${JSON.stringify(['*'])}`
+await $(execaOptions)`${kuboBin} config --json Gateway.HTTPHeaders.Access-Control-Allow-Methods  ${JSON.stringify(['GET', 'POST', 'PUT', 'OPTIONS'])}`
```
SgtPooki, line 65: this file could probably benefit from some optimizations from kubo folks. 

cc @lidel @hacdias

src/pages/redirects-interstitial.tsx
```diff
@@ -1,34 +1,41 @@
...
-    url.searchParams.delete('helia-sw')
-    url.pathname = heliaSw
+  if (!isSubdomainCheckDone) {
+    return (<>First-hit on IPFS hosted service-worker-gateway. Determining state...</>)
   }
-  return url
+
+  if (subdomainRedirectUrl == null) {
+    return <RedirectPage />
+  }
```
SgtPooki: we now render the redirect page if ?helia-sw is rendered, but not by conflating logic into the actual RedirectPage component.

src/pages/redirects-interstitial.tsx
```diff
@@ -1,34 +1,41 @@
...
+
+  return <>Waiting for redirect to subdomain...</>
```
SgtPooki, line 62: if we redirect to a subdomain, this page will not be rendered again for the requested URL... RedirectPage will render directly. This page should also *not render again* for any subsequent unique urls because the SW is registered and will trigger the redirect logic, which would then load RedirectPage if a "first-hit" for that subdomain.

SgtPooki, line 62: I should add these as jsdoc comments

SgtPooki (commented):
self review

SgtPooki:
> Thanks! Had no bandwidth today to review code, but if this data point is useful: I did run it locally against static http server and `?helia-sw=/ipns/docs.ipfs.tech` and it no longer redirects to `/ipns/docs.ipfs.tech` so the infinite loop from [#211 (comment)](https://github.com/ipfs-shipyard/service-worker-gateway/issues/211#issuecomment-2072428579) does not occur anymore.

The http-server hosted scenarios seemed fine. The _redirects logic with the newer `redirects-interstitial.jsx` page introduced the bug. This is why I had to implement local `ipfs-gateway.js` so that we could serve the site similarly to `inbrowser.tld`.

SgtPooki:
merging to fix the bug on https://inbrowser.dev

PR #217, opened 2024-04-23
fix: use separate logger for ui/sw

### LINKED ISSUE #139: Use separate logger in sw and frontend
## Background

When debugging, it's really useful to be able to tell whether the logs are coming from the service worker or the frontend. 

This is a little tricky given that some of the modules in this repo are shared by both the frontend and the service worker. 

## Possible approach

- Use the defaultLogger and call forComponent based on where it's used https://github.com/libp2p/js-libp2p/blob/main/packages/logger/src/index.ts#L171
- For modules that use the logger and are shared by both the SW and FE, we use dep injection to make sure they use a different logger (so as to make it easier to discern where the logs are coming from)

SgtPooki:
- chore: use prefixLogger
- chore: use new logger in sw.ts
- fix: use separate logger for ui/sw

## Title

fix: use separate logger for ui/sw

## Description

Fixes https://github.com/ipfs-shipyard/service-worker-gateway/issues/139

All logs now:

1. include the host.
    - For UI: useful for redirect and config page where things are logged from iframe
    - For SW: useful for debugging different messages from service worker loaded on page in iframe vs main page.
1. have a global prefix: `helia:service-worker-gateway:`
1. have a distinct prefix.
    - For UI: `helia:service-worker-gateway:ui`
    - For SW: `helia:service-worker-gateway:sw`
1. are injected into components that need them

#### filtering console logs by `helia:service-worker-gateway:sw`

image

#### filtering console logs by `helia:service-worker-gateway:ui`

image

## Change checklist

### DIFF
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -1,7 +1,3 @@
-import { logger } from '@libp2p/logger'
+import { prefixLogger } from '@libp2p/logger'
 
-const logObj = logger('helia:service-worker-gateway')
-
-export const log = logObj
-export const error = logObj.error
-export const trace = logObj.trace
+export const logger = prefixLogger('helia:service-worker-gateway')
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -1,6 +1,13 @@
-import { prefixLogger } from '@libp2p/logger'
+import { prefixLogger, type ComponentLogger } from '@libp2p/logger'
 
 const host = globalThis.location.host.replace(':', '_')
-export const swLogger = prefixLogger(`helia:service-worker-gateway:sw:${host}`)
 
-export const uiLogger = prefixLogger(`helia:service-worker-gateway:ui:${host}`)
+const swLogPrefix = `helia:service-worker-gateway:sw:${host}`
+const uiLogPrefix = `helia:service-worker-gateway:ui:${host}`
+
+export const swLogger = prefixLogger(swLogPrefix)
+export const uiLogger = prefixLogger(uiLogPrefix)
+
+export const getUiComponentLogger = (component: string): ComponentLogger => {
+  return prefixLogger(`${uiLogPrefix}:${component}`)
+}
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -1,3 +1,6 @@
 import { prefixLogger } from '@libp2p/logger'
 
-export const logger = prefixLogger('helia:service-worker-gateway')
+const host = globalThis.location.host.replace(':', '_')
+export const swLogger = prefixLogger(`helia:service-worker-gateway:sw:${host}`)
+
+export const uiLogger = prefixLogger(`helia:service-worker-gateway:ui:${host}`)
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -2,8 +2,8 @@ import { prefixLogger, type ComponentLogger } from '@libp2p/logger'
 
 const host = globalThis.location.host.replace(':', '_')
 
-const swLogPrefix = `helia:service-worker-gateway:sw:${host}`
-const uiLogPrefix = `helia:service-worker-gateway:ui:${host}`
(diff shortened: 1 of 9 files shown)


src/pages/redirect-page.tsx
```diff
@@ -27,7 +29,7 @@ const ConfigIframe = (): React.JSX.Element => {
   )
 }
 
-const channel = new HeliaServiceWorkerCommsChannel('WINDOW')
+const channel = new HeliaServiceWorkerCommsChannel('WINDOW', uiLogger)
```
SgtPooki: without modifying libp2p, we probably need to export some kind of `getUiComponentLogger` from `lib/logger.ts` so we know this `HeliaServiceWorkerCommsChannel` is from the `redirect-page`.

libp2p logger functions are nice for a single level of components, but don't seem to support multiple levels of nesting

src/pages/redirect-page.tsx
```diff
@@ -41,14 +43,14 @@ function RedirectPage ({ showConfigIframe = true }: { showConfigIframe?: boolean
 
     async function doWork (config: ConfigDb): Promise<void> {
       try {
-        await setConfig(config)
+        await setConfig(config, uiLogger)
```
SgtPooki: we should be able to tell this is from redirect-page.. will update.

src/sw.ts
```diff
@@ -322,16 +323,16 @@ async function fetchAndUpdateCache (event: FetchEvent, url: URL, cacheKey: strin
...
+    log.trace('response headers: %s: %s', key, value)
   })
 
-  log('helia-sw: response status: %s', response.status)
+  log('response status: %s', response.status)
 
   try {
     await storeReponseInCache({ response, isMutable: true, cacheKey, event })
-    trace('helia-ws: updated cache for %s', cacheKey)
+    log.trace('helia-ws: updated cache for %s', cacheKey)
```
SgtPooki: need to remove this "helia-ws" typo msg prefix.

src/sw.ts
```diff
@@ -141,7 +142,7 @@ self.addEventListener('activate', (event) => {
       case 'RELOAD_CONFIG':
         void updateVerifiedFetch().then(async () => {
           channel.postMessage<any>({ action: 'RELOAD_CONFIG_SUCCESS', data: { config: await getConfig() } })
-          trace('sw: RELOAD_CONFIG_SUCCESS for %s', self.location.origin)
+          log.trace('sw: RELOAD_CONFIG_SUCCESS for %s', self.location.origin)
```
SgtPooki: need to remove "sw:" prefix

SgtPooki (commented):
self review. updates incoming.

src/lib/logger.ts
```diff
@@ -1,7 +1,13 @@
-import { logger } from '@libp2p/logger'
+import { prefixLogger, type ComponentLogger } from '@libp2p/logger'
 
-const logObj = logger('helia:service-worker-gateway')
+const host = globalThis.location.host.replace(':', '_')
 
-export const log = logObj
-export const error = logObj.error
-export const trace = logObj.trace
+const swLogPrefix = `helia:service-worker-gateway:sw:${host}`
```
2color:
```suggestion
const swLogPrefix = `helia:sw-gateway:sw:${host}`
```

Any thoughts on making the logs a little shorter since we use sw as an acception abbreviation elsewhere too?

2color (approved):
Really nice!

src/lib/logger.ts
SgtPooki: Yeaaa i thought about that. Theyre pretty long, so if you think shorter is fine im absolutely a fan.

I think uniqueness is the most important here, instead of verbosity. If someone is looking at these logs, theyre likely aware its a "service-worker-gateway"

SgtPooki: Is "helia:gateway:sw" & "helia:gateway:ui" too obtuse?

PR #190, opened 2024-04-10
feat: migrate to preact/compat

SgtPooki:
## Title

feat: migrate to preact/compat

## Description

Built on top of https://github.com/ipfs-shipyard/service-worker-gateway/pull/188 intended to test migration to preact. **The size reduction is 0.5M**

## Notes & open questions

Things work fine, but preact/compat says suspense and lazy are experimental and shouldn't be used in prod.

#### Size and assets on `fix/dynamic-imported-ui-components`

```bash
total 4.7M
drwxr-xr-x 33 sgtpooki 1.1K Apr 10 13:41 ..
-rw-r--r--  1 sgtpooki  30K Apr 10 13:41 ipfs-sw-778.js
-rw-r--r--  1 sgtpooki 5.2K Apr 10 13:41 ipfs-sw-778.css
-rw-r--r--  1 sgtpooki  51K Apr 10 13:41 ipfs-sw-main.js
-rw-r--r--  1 sgtpooki  16K Apr 10 13:41 ipfs-sw-781.js
-rw-r--r--  1 sgtpooki  25K Apr 10 13:41 ipfs-sw-850.js
-rw-r--r--  1 sgtpooki 2.5K Apr 10 13:41 ipfs-sw-992.js
-rw-r--r--  1 sgtpooki 2.5K Apr 10 13:41 ipfs-sw-611.js
-rw-r--r--  1 sgtpooki  55K Apr 10 13:41 ipfs-sw-sw.js
-rw-r--r--  1 sgtpooki 2.7K Apr 10 13:41 ipfs-sw-styles.css
-rw-r--r--  1 sgtpooki  57K Apr 10 13:41 ipfs-sw-117.js
-rw-r--r--  1 sgtpooki  223 Apr 10 13:41 ipfs-sw-styles.js
-rw-r--r--  1 sgtpooki  88K Apr 10 13:41 ipfs-sw-96.js
-rw-r--r--  1 sgtpooki   26 Apr 10 13:41 _redirects
-rw-r--r--  1 sgtpooki 577K Apr 10 13:41 ipfs-sw-vendor-react.js
-rw-r--r--  1 sgtpooki 3.4M Apr 10 13:41 ipfs-sw-561.js
-rw-r--r--  1 sgtpooki 6.1K Apr 10 13:41 ipfs-sw-117.js.LICENSE.txt
drwxr-xr-x 31 sgtpooki  992 Apr 10 13:41 .
[...]

PR #219, opened 2024-04-24
fix: revert "feat: migrate to preact/compat (#190)"

SgtPooki:
## Title

fix: revert "feat: migrate to preact/compat (#190)"

## Description

This reverts commit 8c98f579b95992072474f02fe39865fda825379d from PR #190 

## Notes & open questions

I believe this was causing some issues in deployed environments.. I was also able to reproduce some errors locally with `http-server` and `npm start`.. Issues include:

1. not being able to find css/js assets consistently
2. CMD+SHIFT+R resulting in renderable content, CMD+R resulting in missing assets and failed loading
3. css file not found for config page when attempting to load subdomain config page directly (found when trying to look into https://github.com/ipfs-shipyard/service-worker-gateway/issues/218)

Apparently preact support is not quite where we need it with webpack 5, or we need to do some additional config. Considering the large reduction in final asset size, it was worth a try, but  reverting now because potential pains & [incompatibility](https://github.com/preactjs/preact-cli/issues/1579) are not worth it. 

We can potentially investigate this further later, but we would need some way to reproduce these errors consistently and write some test to catch those failures.

## Change checklist

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -14,7 +14,7 @@ export default {
   dependencyCheck: {
     ignore: [
       // .jsx files aren't checked properly.
-      'preact',
+      'react-dom',
 
       // required by webpack
       'webpack-cli',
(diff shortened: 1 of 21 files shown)


SgtPooki (commented):
self review

PR #233, opened 2024-04-29
feat: _redirects only for /ipns and /ipfs

### LINKED ISSUE #223: feat: optimize _redirects to only handle splats of /ipfs and /ipns requests
@2color reported getting the following error:

> Still not working, the main problem is the one I mentioned where requests for js assets, e.g. https://inbrowser.dev/ipfs-sw-117-78590e7d2e2caa94e1bc.js are redirected to /?helia-sw=/ipfs-sw-117-78590e7d2e2caa94e1bc.js and then that returns the root html page

We should make sure we're not redirecting these which can cause even more confusion

SgtPooki:
## Title

feat: _redirects only for /ipns and /ipfs

## Description

Fixes https://github.com/ipfs-shipyard/service-worker-gateway/issues/223

## Notes & open questions

I don't think this will cause unexpected errors but cc @lidel for double checking

## Change checklist

### DIFF
--- a/public/_redirects
+++ b/public/_redirects
@@ -1,1 +1,2 @@
-/* /?helia-sw=/:splat 302
+/ipns/* /?helia-sw=/ipns/:splat 302
+/ipfs/* /?helia-sw=/ipfs/:splat 302


2color (approved):
Looks right to me.

PR #234, opened 2024-04-29
feat: cache sw assets with service worker

### LINKED ISSUE #222: feat: cache sw-gateway assets
Instead of ignoring `ipfs-sw-*` assets, we could let the fetch request go through, and cache it, so unstable servers (inbrowser) are less burdened with requests for all the assets.

However, this could add more issues with caching.. but all assets have content-id-hashes, so a sw caching all the assets for it's deployment should be fine. open to thoughts

SgtPooki:
## Title

feat: cache sw assets with service worker

## Description

Fixes https://github.com/ipfs-shipyard/service-worker-gateway/issues/222

This PR enables caching of `ipfs-sw-*` assets with the service worker. This will allow the service worker to cache the assets and serve them from the cache when the user is offline.
It will also allow users to load assets more quickly, and since all assets have a postfixed `contenthash` in their filename, we shouldn't have to worry about stale assets being served.

Also FYI: requests for `ipfs-sw-sw.js` (the root service worker that is registered) will *not* hit this cache, because service workers can't see requests for themselves.

## Notes & open questions

This enables offline browsing of sites. Before this change, root assets would not be accessible if user is offline:

### Offline with changes in `main`

https://github.com/ipfs-shipyard/service-worker-gateway/assets/1173416/8ff74c8c-f141-445e-ab4d-634e6382e377

### Offline with changes from this PR

https://github.com/ipfs-shipyard/service-worker-gateway/assets/1173416/777d6de8-1ee6-4ca0-8e80-afdbf8be79c6

## Change checklist


SgtPooki (commented):
self review

src/sw.ts
```diff
@@ -285,7 +301,15 @@ function isAggregateError (err: unknown): err is AggregateError {
...
 function isSwAssetRequest (event: FetchEvent): boolean {
   const isActualSwAsset = /^.+\/(?:ipfs-sw-).+$/.test(event.request.url)
-  return isActualSwAsset
+  // if path is not set, then it's a request for index.html which we should consider a sw asset
+  const url = new URL(event.request.url)
+  // but only if it's not a subdomain request (root index.html should not be returned for subdomains)
+  const isIndexHtmlRequest = url.pathname === '/' && !isSubdomainRequest(event)
+
+  // if request is for https://cdn.jsdelivr.net/* we should consider it a sw asset
+  const isJsDelivrRequest = /^https:\/\/cdn.jsdelivr.net\/.+$/.test(event.request.url)
```
2color: Tangentially related: what do you think about publishing the CSS as part of this app instead of relying on a third-party origin (jsdeliver).

With both Cloudflare and the SW Cache, should be performant enough.

src/sw.ts
```diff
@@ -199,8 +200,23 @@ async function requestRouting (event: FetchEvent, url: URL): Promise<boolean> {
...
+    /**
+     * Return the asset from the cache if it exists, otherwise fetch it.
+     */
+    event.respondWith(caches.open(CURRENT_CACHES.swAssets).then(async (cache) => {
+      const cachedResponse = await cache.match(event.request)
+      if (cachedResponse != null) {
+        return cachedResponse
+      }
+      const response = await fetch(event.request)
+      await cache.put(event.request, response.clone())
```
2color, line 218: One thing to consider is that these won't be evicted and may gradually exhaust the storage quota with every release.

src/sw.ts
SgtPooki: Yea i started to do that previously and i think i squashed it because of other changes. I could do it as a part of this pr

src/sw.ts
SgtPooki: We could clear swAssets on "install" just to be safe

src/sw.ts
SgtPooki: now that i'm at my laptop, I think i started to do that with the dynamic imports, but it only made the bundle larger.. now that index.ts and app.tsx is smaller, and we have a router, I could include tachyons and ipfs-css on only helper-ui, redirect-page, and config, and it will reduce burden for redirect-interstitial (ipfs first-hit without sw registered)

src/sw.ts
SgtPooki: just confirmation for posterity, install is only fired once per "service worker instance." See https://web.dev/articles/service-worker-lifecycle#install_2

SgtPooki: fixed by [`a6c3305` (#234)](https://github.com/ipfs-shipyard/service-worker-gateway/pull/234/commits/a6c330534c0c03525ccf1f009638182c1ceaf7b5)

src/sw.ts
SgtPooki: fixed by [`20a9c6d` (#234)](https://github.com/ipfs-shipyard/service-worker-gateway/pull/234/commits/20a9c6d0f92a4dbf63d556d062b52826c1f2627f)

note that because the bundle is larger now (`-rw-r--r--  1 sgtpooki  13M Apr 30 08:08 ipfs-sw-138-66da405c07a0317cdbbf.css`), there is FOUC on routing determination.

I don't think we should include it in the bundle.. but also, I think we should because then we're not dependent upon a third party.. ¯\\\_(ツ)\_/¯

SgtPooki: @lidel could use eyes on this one here.

PR #258, opened 2024-05-14
feat: add content-loading indication page

### LINKED ISSUE #246: UX: visual confirmation that load is in progress
## Problem 

Right now, when user clicks on "Load content" and if trustless gateway is slow, it just hangs like that. 

It looks like things are broken with SW, and not IPFS retrieval, but if you inspect log, SW does its job, it has block request in pending state to render `/wiki/` directory: 

> 

It is stuck like this for  4.8 min and then errors:

> image

## Solution

We need visual indicator that page load started.

At minimum, clicking on "Load content" should replace page with message "SW initialized, please wait while it attempts to retrieve content from IPFS peers"

Over time, the page would display additional tips, if the load takes longer than a few seconds, but right now we need basic UX fix.

SgtPooki:
## Title

feat: add content-loading indication page

## Description

Fixes https://github.com/ipfs/service-worker-gateway/issues/246

## Notes & open questions

New loading page that displays when content should be loading to handle scenarios where fetching content takes a while

## Change checklist


SgtPooki (commented):
self review

lidel (approved):
Thanks! Lgtm, but was missing some animation to distract user for a few seconds while things load. We already had heartbeat splash in ipfs-desktop/companion, so reused it (the cube below acts as splash).

Preview:

> 
> Screen Shot 2024-05-15 at 22 31 09

PR #263, opened 2024-05-20
feat: about section on subdomain loads

lidel:
This PR moves "about" into separate component and reuses it on subdomain landing page displayed when SW is not registered yet.

This aims to fix UX problem when I send https://en-wikipedia--on--ipfs-org.ipns.inbrowser.dev to someone, and  all they see is no wikipedia, but a confusing button to click, and some configuration, but have no idea what is happening.

Feedback welcome, preview below:


SgtPooki (approved):
awesome

PR #282, opened 2024-06-05
fix: remove fonts except woff2

SgtPooki:
## Title

fix: remove fonts except woff2

## Description

Fixes https://github.com/ipfs/service-worker-gateway/issues/281

### before change 

`du -ch dist/*.css dist/*.js | grep total` = **18M**

```sh
╰─ ✔ ❯ ll dist/*.css dist/*.js
-rw-r--r-- 1 sgtpooki  63K Jun  5 10:39 dist/ipfs-sw-sw.js
-rw-r--r-- 1 sgtpooki    2 Jun  5 10:39 dist/ipfs-sw-781-fbb750fd312778403036.css
-rw-r--r-- 1 sgtpooki 6.1K Jun  5 10:39 dist/ipfs-sw-51-dfae7cbe5a15c9f08cba.css
-rw-r--r-- 1 sgtpooki 5.2K Jun  5 10:39 dist/ipfs-sw-56-dc9b2c9958c23bc8fff0.css
-rw-r--r-- 1 sgtpooki 6.1K Jun  5 10:39 dist/ipfs-sw-776-e5915cc60e3552ecbd7c.css
-rw-r--r-- 1 sgtpooki 2.3K Jun  5 10:39 dist/ipfs-sw-styles-bd392c278845011bf0d3.css
-rw-r--r-- 1 sgtpooki  58K Jun  5 10:39 dist/ipfs-sw-main-544ab9e12c39034c7da2.js
-rw-r--r-- 1 sgtpooki  13M Jun  5 10:39 dist/ipfs-sw-138-66da405c07a0317cdbbf.css
-rw-r--r-- 1 sgtpooki  17K Jun  5 10:39 dist/ipfs-sw-56-806019a37f15985d8a2b.js
-rw-r--r-- 1 sgtpooki  15K Jun  5 10:39 dist/ipfs-sw-776-5a14701760f367fc0b71.js
-rw-r--r-- 1 sgtpooki  37K Jun  5 10:39 dist/ipfs-sw-781-756e5b0b09ef3eedde5e.js
-rw-r--r-- 1 sgtpooki  17K Jun  5 10:39 dist/ipfs-sw-51-2198ef777aa27dc13ed0.js
-rw-r--r-- 1 sgtpooki 2.8K Jun  5 10:39 dist/ipfs-sw-992-0a574e46705faad5f353.js
[...]

PR #293, opened 2024-06-11
chore: improve LCP performance

### LINKED ISSUE #281: CSS bundle is huge
## Problem

One of the CSS files that is loaded in the bundle is pretty big (~12MiB and 9MiB gzipped)

Specifically it's `https://inbrowser.link/ipfs-sw-138-66da405c07a0317cdbbf.css`

A look at the build reveals that this is coming from the `ipfs-css` module which we import: https://github.com/ipfs/service-worker-gateway/actions/runs/9167097012/job/25203686469#step:4:98 

The actual css file is much smaller, https://github.com/ipfs-shipyard/ipfs-css/blob/main/ipfs.css is around 18.5 kb, but because the imported typefaces are inlined as base64, it makes the css bundle explode. 

The main problem is that most of the bundled typefaces are not even used by the frontend. Moreover, we load three formats for each typeface (woff, woff2, otf)

## Possible solutions:

- Add purgecss to purge any unused css https://purgecss.com/plugins/webpack.html
- Remove older typeface formats from the bundle, i.e. woff and otf. The only impact this will have is visual — but who cares. People are trying to load a CID/Dapps. This won't be a major regression.

2color:
## Description 

This PR builds on #282 and improves the LCP performance.

Attempts to fix #281 

## What and Why

- Disable inlining of fonts into the css bundle. Reduces the build size significantly and results in only the needed fonts being loaded. As a result of this change, the font files are copied to the dist folder, but **are not prefixed** with `ipfs-sw`, so we may have trouble if the fonts need to be loaded once the SW is installed. 
- Disable inlining of the sourcemaps for production builds. This one results in significant reduction in the build size. The source maps are also prefixed with `ipfs-sw` so there shouldn't be any major problems.

## Notes & open questions

- how to prefix the downloaded/copied typefaces so that they can work with the service worker 
- Should we just remove use of external fonts and use [system-ui](https://css-tricks.com/snippets/css/system-font-stack/)? 
- We could also use the `font-display` css property to `optional` to ensure we only use fonts if fetching them takes less than 100ms. (https://web.dev/articles/font-best-practices#choose_an_appropriate_font-display_strategy)

## Change checklist

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 import ReactDOMClient from 'react-dom/client'
 import App from './app.jsx'
 import { RouterProvider, type Route } from './context/router-context.jsx'
+import * as renderChecks from './lib/routing-render-checks.js'
 
 // SW did not trigger for this request
 const container = document.getElementById('root')
@@ -19,7 +20,6 @@ const routes: Route[] = [
   { path: '#/ipfs-sw-config', shouldRender: async () => (await import('./lib/routing-render-checks.js')).shouldRenderConfigPage(), component: LazyConfig },
   {
     shouldRender: async () => {
-      const renderChecks = await import('./lib/routing-render-checks.js')
       return renderChecks.shouldRenderRedirectPage()
     },
     component: LazyRedirectPage
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -16,12 +16,10 @@ const LazyInterstitial = React.lazy(async () => import('./pages/redirects-inters
 
 const routes: Route[] = [
   { default: true, component: LazyHelperUi },
-  { shouldRender: async () => (await import('./lib/routing-render-checks.js')).shouldRenderRedirectsInterstitial(), component: LazyInterstitial },
-  { path: '#/ipfs-sw-config', shouldRender: async () => (await import('./lib/routing-render-checks.js')).shouldRenderConfigPage(), component: LazyConfig },
+  { shouldRender: async () => renderChecks.shouldRenderRedirectsInterstitial(), component: LazyInterstitial },
+  { path: '#/ipfs-sw-config', shouldRender: async () => renderChecks.shouldRenderConfigPage(), component: LazyConfig },
   {
-    shouldRender: async () => {
-      return renderChecks.shouldRenderRedirectPage()
-    },
+    shouldRender: async () => renderChecks.shouldRenderRedirectPage(),
     component: LazyRedirectPage
   }
 ]
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -2,8 +2,10 @@ import { execSync } from 'child_process'
 import path from 'node:path'
 import { fileURLToPath } from 'node:url'
 import CopyWebpackPlugin from 'copy-webpack-plugin'
+import { sync as globSync } from 'glob'
 import HtmlWebpackPlugin from 'html-webpack-plugin'
 import MiniCssExtractPlugin from 'mini-css-extract-plugin'
+import { PurgeCSSPlugin } from 'purgecss-webpack-plugin'
 import TerserPlugin from 'terser-webpack-plugin'
 import webpack from 'webpack'
 import BundleAnalyzerPlugin from 'webpack-bundle-analyzer'
@@ -234,6 +236,9 @@ const common = {
     new MiniCssExtractPlugin({
       filename: 'ipfs-sw-[name]-[contenthash].css',
       chunkFilename: 'ipfs-sw-[id]-[contenthash].css'
+    }),
+    new PurgeCSSPlugin({
+      paths: globSync(`${paths.distTsc}/**/*`, { nodir: true })
     })
   ],
 
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -272,7 +272,23 @@ const common = {
         test: /\.(sa|sc|c)ss$/i,
         use: [
           MiniCssExtractPlugin.loader,
-          'css-loader'
+          {
+            loader: 'css-loader',
+            options: {
+              url: {
+                // Exclude fonts from build
+                filter: (url, resourcePath) => {
+                  const fontRegex = /\.(otf|woff|woff2)/
+
+                  if (url.search(fontRegex)) {
+                    return false
+                  }
+
+                  return true
+                }
+              }
+            }
+          }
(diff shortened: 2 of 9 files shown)


webpack.config.js
```diff
@@ -266,8 +266,8 @@ const common = {
       // Images: Copy image files to build folder
       { test: /\.(?:ico|gif|png|jpg|jpeg)$/i, type: 'asset/resource' },
 
-      // Fonts and SVGs: Inline files
-      { test: /\.(woff2|svg|)$/, type: 'asset/inline' },
+      // Inline SVGs
+      { test: /\.(svg|)$/, type: 'asset/inline' },
```
SgtPooki, line 270: I'm wondering if the fonts are still in the dist folder. will peek before approving/merging

webpack.config.js
```diff
@@ -60,7 +60,7 @@ const paths = {
  */
 const prod = {
   mode: 'production',
-  devtool: 'inline-source-map',
+  devtool: 'source-map',
```
SgtPooki, line 63: +1. this may cause some issues with debugging, especially because of our currently convoluted build process (aegir/tsc->webpack) but will definitely reduce prod bundle size.

SgtPooki:
```bash
╰─ ✔ ❯ ls -lhatr dist                                                                                                                5.98   27.5G   65%   100%  ─╯
Alias tip: ll dist
total 14M
drwxr-xr-x  38 sgtpooki 1.2K Jun 12 11:34 ..
-rw-r--r--   1 sgtpooki 132K Jun 12 11:34 81862cd5e78843a135df.woff
-rw-r--r--   1 sgtpooki 139K Jun 12 11:34 e19bfcb1c7eac9952c86.woff
-rw-r--r--   1 sgtpooki 103K Jun 12 11:34 a41f98d0e72e4b39f67a.woff2
-rw-r--r--   1 sgtpooki 145K Jun 12 11:34 11788b15932bb4e5cae6.woff
-rw-r--r--   1 sgtpooki 244K Jun 12 11:34 a09fa2ba55dff7af7fc0.otf
# ...
-rw-r--r--   1 sgtpooki 110K Jun 12 11:34 551a291434a17555ee24.eot
```

looking at the dist folder, a few issues:

1. we're still including a lot of fonts that I don't think we need. 
2. assets are not prefixed with `ipfs-sw` which may cause issues with re-requests or SW handling of them


--- 

addressing items in your notes section:

> how to prefix the downloaded/copied typefaces so that they can work with the service worker

Just saw this. you were already aware.. I should have known ;)

> Should we just remove use of external fonts and use [system-ui](https://css-tricks.com/snippets/css/system-font-stack/)?

If this doesn't require an update to ipfs-css, I am down. I like the idea of reducing dependencies

> We could also use the font-display css property to optional to ensure we only use fonts if fetching them takes less than 100ms. ([web.dev/articles/font-best-practices#choose_an_appropriate_font-display_strategy](https://web.dev/articles/font-best-practices#choose_an_appropriate_font-display_strategy))

From that page: 
> The block period begins when the browser requests a web font. During the block period, if the web font is not available, the font is rendered in an invisible fallback font and thus the text will not be visible to the user. If the font is not available at the end of the block period, it will be rendered in the fallback font.

in other words, a flash of unrendered content(FOUC) for 100ms for any requested fonts. this could be undesirable if we aren't caching any font assets with SW.

I think I prefer a system-ui font

webpack.config.js
2color: Why do we use both aegir and webpack for this repo. Aegir comes with esbuild, shouldn't esbuild serve us?

2color:
> +1. this may cause some issues with debugging, especially because of our currently convoluted build process (aegir/tsc->webpack) but will definitely reduce prod bundle size.

I'm curious, why do we use both aegir (which comes with esbuild) and webpack? Is that because of the code splitting?

2color:
> we're still including a lot of fonts that I don't think we need.

[Fixed in the last build which does not copy any of the fonts from `url` functions in the css](https://github.com/ipfs/service-worker-gateway/pull/293/commits/b0d71086fe3811cbf5c94f41ae3c33b8d0188483#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fR281)

> I think I prefer a system-ui font

That's what I've defaulted to now by overriding the `.sans-serif` class that's [defined in ipfs-css](https://github.com/ipfs-shipyard/ipfs-css/blob/main/ipfs.css#L276). 

So no more external fonts.

webpack.config.js
2color: Yep. But removed in the last build.

2color:
This PR is ready for another review. 

This still doesn't change the fact that CSS files are getting chunked in a slightly weird way (multiple small files probably chunked based on the dynamic imports), and are not minified.

Compared to 947011e772f73d32b2380bb6d7521f61fc8d27bf there are some big savings:

### Before
Screenshot 2024-06-13 at 3 47 35 pm
### After
Screenshot 2024-06-13 at 3 47 01 pm

SgtPooki:
> > +1. this may cause some issues with debugging, especially because of our currently convoluted build process (aegir/tsc->webpack) but will definitely reduce prod bundle size.
> 
> I'm curious, why do we use both aegir (which comes with esbuild) and webpack? Is that because of the code splitting?

Dynamic imports and benefits of asset loading and sw stuff I'm not as familiar with esbuild for. I attempted a migration fully to esbuild in the past but it didn't work great.

src/index.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import ReactDOMClient from 'react-dom/client'
 import App from './app.jsx'
 import { RouterProvider, type Route } from './context/router-context.jsx'
+import * as renderChecks from './lib/routing-render-checks.js'
```
SgtPooki, line 5: i'll push a change up updating this for the rest of the imports.. it would be cleaner in this file.

SgtPooki: pushed up [`b011e46` (#293)](https://github.com/ipfs/service-worker-gateway/pull/293/commits/b011e46ac4bffa68ac972d177b4cc40af309ff1a)

SgtPooki (approved):

src/index.tsx
2color: Nice! I totally missed that.

PR #294, opened 2024-06-11
ci: add lighthouse checks to ci

2color:
This PR adds a new GH Actions Job to run Lighthouse checks against the build outputs.

In its current form, it runs a static server which serves the build and runs Lighthouse tests against it. This means that it doesn't resemble the production deployment exactly (which I don't have enough information about).

## alternative paths

- We could instead run it against inbrowser.link upon releases and update the commit status so that we can correlate lighthouse results to changes in the code.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,20 @@ jobs:
     - run: npm run --if-present dep-check
     - run: npm run --if-present doc-check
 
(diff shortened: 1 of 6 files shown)


SgtPooki:
We should update this to run after the "deploy to staging" and "deploy to prod" workflows against the published endpoint

Issue #289, opened 2024-06-10
ux: unable to exit config editor on inbrowser.tld


lidel:
## Problem

If I open https://inbrowser.dev and click on  config editor:

> ![2024-06-10_15-25](https://github.com/ipfs/service-worker-gateway/assets/157609/b7009a56-b55e-46c5-a54e-eb2ad9f1c10a)

There is no way to close editor and go back to the main screen where i can enter the content path, i am stuck in config editor and have to manually edit URL in address bar:

> ![2024-06-10_15-22](https://github.com/ipfs/service-worker-gateway/assets/157609/2b54d4f7-6a73-49b2-bcca-6da1708eabc2)

## Fix ideas

- make clickin "Save config" collapse config when embedded on Iframe and remove `/#/ipfs-sw-config` from URL if not in iframe (on main domain)
- Or keep "Save Config" as-is and add dedicated "Close" button that does the above.

SgtPooki:
> make clicking "Save config" collapse config when embedded on Iframe and remove /#/ipfs-sw-config from URL if not in iframe (on main domain)

this is supposed to already exist. Will look into this now.

SgtPooki:
~~"save config" does close config edit screen on inbrowser.dev and local.~~ 

Actually, after clicking "reset config" it doesn't... interesting.

SgtPooki:
> Actually, after clicking "reset config" it doesn't... interesting.

After clearing all storage in the browser, and having a valid config, the reload button works fine, I can't reproduce on inbrowser.dev when using the settings gear icon directly from the homepage, and clicking "save config" to close it.


clicking "reset config" and then "save config" also continues to work after I cleared browser settings

SgtPooki:
and now I can reproduce locally with the code from https://github.com/ipfs/service-worker-gateway/pull/293.... I need to figure out what is going on

SgtPooki:
This was fixed by https://github.com/ipfs/service-worker-gateway/commit/c83830714475a7a20d949d5bf9f74bf65c7e387d & the implementation of `await fetch('/#/ipfs-sw-config-reload')`

PR #216, opened 2024-04-23
chore(main): release 1.2.0

lidel:
:robot: I have created a release *beep* *boop*
---

### Features

* _redirects only for /ipns and /ipfs ([#233](https://github.com/ipfs/service-worker-gateway/issues/233)) ([89faa58](https://github.com/ipfs/service-worker-gateway/commit/89faa588ee0c033f3d2631d5c2823c005f17ead5))
* add content-loading indication page ([#258](https://github.com/ipfs/service-worker-gateway/issues/258)) ([c815a25](https://github.com/ipfs/service-worker-gateway/commit/c815a25d2657244cde5cea95cf535dd1bc728675))
* cache sw assets with service worker ([#234](https://github.com/ipfs/service-worker-gateway/issues/234)) ([20a8f32](https://github.com/ipfs/service-worker-gateway/commit/20a8f329402dac4cba53fec55bd059eeb7f7137d))
* landing & config page UX improvements ([#235](https://github.com/ipfs/service-worker-gateway/issues/235)) ([fb9b04e](https://github.com/ipfs/service-worker-gateway/commit/fb9b04e06b6b3670728c243838ee86f3d6b55067))
* meaningful error pages ([#195](https://github.com/ipfs/service-worker-gateway/issues/195)) ([80774f5](https://github.com/ipfs/service-worker-gateway/commit/80774f50760e4b757dfacfb7d5df79f3fb020012))

### Bug Fixes

* ipfs-hosted redirects are not infinite ([#215](https://github.com/ipfs/service-worker-gateway/issues/215)) ([40cc8c7](https://github.com/ipfs/service-worker-gateway/commit/40cc8c7612302e60a2be2c9372e22e95fa10a799))
* multiple bugs ([#220](https://github.com/ipfs/service-worker-gateway/issues/220)) ([75aa0b8](https://github.com/ipfs/service-worker-gateway/commit/75aa0b8773c8f1ab484ac110fe607404303e2e8c))
* revert "feat: migrate to preact/compat ([#190](https://github.com/ipfs/service-worker-gateway/issues/190))" ([#219](https://github.com/ipfs/service-worker-gateway/issues/219)) ([4443174](https://github.com/ipfs/service-worker-gateway/commit/44431745be1f5638fb47969326865c8825ade141))
[...]

PR #295, opened 2024-06-12
chore: update staging

SgtPooki:
updates staging with the following commits from main:

- chore(main): release 1.2.0 (#216)
- feat: about section on subdomain loads (#263)
- fix: remove fonts except woff2 (#282)

---

Mostly trying this because I can't repro https://github.com/ipfs/service-worker-gateway/issues/289 on localhost.. but I don't see changes here that could be breaking things on inbrowser.dev :(


2color:
How do you create these releases?

SgtPooki:
> How do you create these releases?

main is supposed to be our dev branch, and staging is not sharing history with main, so I just did:

```sh
git checkout staging
git pull
git checkout main
git pull
gh pr create --base staging
```

2color (approved):

PR #301, opened 2024-06-18
feat: probe and use local gateway if available

### LINKED ISSUE #299: ux: detect local trustless gw and default to it

2color:
## Description

This PR checks if there's a local trustless gateway available (from IPFS Desktop or Kubo running on the user's machine) and adds it to the configuration. 

Fixes #299

Adding it to the config on load is convoluted due to the way we persist and pass config from across origins into service workers.

## Notes & open questions

- In all my tests, I haven't been able to get the service worker to actually make requests to the local gateway. I couldn't find any such restriction in the [service worker docs](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API). 

## Change checklist

### DIFF
--- a/src/lib/local-gateway.ts
+++ b/src/lib/local-gateway.ts
@@ -1,0 +1,32 @@
+import { uiLogger } from './logger'
+
+export const localGwUrl = 'http://127.0.0.1:8080'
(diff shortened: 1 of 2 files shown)


lidel:
Continued in https://github.com/ipfs/service-worker-gateway/pull/303

Issue #288, opened 2024-06-10
ux: add configuration toggle for delegated routing


lidel:
> IPFS Camp UX improvement suggestion based on discussions with @aschmahmann  

## Current state 

- `service-worker-gateway` sends a lot of requests to `https://delegated-routing.dev` but (afaik) we don't have direct retrieval yet, so there is no benefit from it

## Proposed UX fix

- Empty list of deleegated routing endpoints is bad UX, as user needs to know URL of endpoint to enable it
- Let's add a toggle that enables/disables use of addresses defined in the routing configuration field
  - It should be disabled by default, until we have direct retrieval working

aschmahmann:
@lidel I don't know if that fix is quite enough. If the delegated routing endpoints are being used for IPNS resolution (as opposed to using the trustless gateway endpoints) then we still need them around even if not doing direct retrievals.

We could at the Helia level not turn on the content or peer routing systems though (or potentially use the trustless gateways for IPNS fetch).

lidel:
Yes, the idea here is to have "quality of life" toggle that disables all delegated routing, and only relies on trustless gateways.

You can still do IPNS via trustless gateways alone, but if verified-fetch does not leverage it yet, we can simply error on IPNS errors, informing user they need to enable delegated routing.

2color:
This seems like a duplicate of #267. 

>@lidel I don't know if that fix is quite enough. If the delegated routing endpoints are being used for IPNS resolution (as opposed to using the trustless gateway endpoints) then we still need them around even if not doing direct retrievals.

That's right. Currently, Helia is built in a way that accepts 3 kinds of routers (libp2p, delegatedHTTPRouting, and httpGatewayRouting) and by default includes both `delegatedHTTPRouting`, and `httpGatewayRouting`. The problem is that if we remove the delegatedHTTPRouting, there's no way for Helia to Resolve IPNS names. 


I've opened https://github.com/ipfs/helia/issues/558 to address this gap in Helia.

lidel:
Descoping from Camp milestone due to impact on DNSLink (see https://github.com/ipfs/service-worker-gateway/pull/303#pullrequestreview-2148664363).

SgtPooki:
Do we still want to do this? I think the only action item here would be to disable delegated routing if direct retrieval is not enabled and trustless gateways *are* enabled.

achingbrain:
[...]

PR #303, opened 2024-06-19
feat: make routers optional and enable local insecure gateways

### LINKED ISSUE #299: ux: detect local trustless gw and default to it

2color:
## Added by this PR:
- Ability to disable delegated routing using a toggle. When disabled, the `routers` configuration will not be passed to Helia.
- Probe for local gateway and add the local gateway to the configuration if available. If unavailable, remove the local gateway from the configuration. 

Fixes #299 #288 

## Change checklist


src/lib/local-gateway.ts
```diff
@@ -0,0 +1,33 @@
+import { uiLogger } from './logger.js'
+
+export const localGwUrl = 'http://127.0.0.1:8080'
```
lidel: Since we know it is not subdomains fault, we may want to switch to `localhost` which is Secure Context:

```suggestion
export const localGwUrl = 'http://localhost:8080'
```

Should be fine, as long  `fetch` follows redirects by default, or is there more to it?

src/lib/config-db.ts
```diff
@@ -12,7 +12,7 @@ export interface ConfigDb extends BaseDbConfig {
 }
 
 export const defaultGateways = ['https://trustless-gateway.link']
-export const defaultRouters = ['https://delegated-ipfs.dev']
+export const defaultRouters = []
```
lidel: If we remove default delegated router, won't we lose opportunistic  direct retrieval over HTTP? Maybe keeping the default as-is is better play for Camp?

Rationale: 

It seems to work out-of-the-box thanks to session changes in Helia a while ago. Right now it is mostly web3.storage, but a lot of data is there.

You can test with `bafkreibm6jg3ux5qumhcn2b3flc3tyu6dmlb4xa7u5bf44yegnrjhc4yeq`

Example: https://delegated-ipfs.dev/routing/v1/providers/bafkreibm6jg3ux5qumhcn2b3flc3tyu6dmlb4xa7u5bf44yegnrjhc4yeq  returns `/dns/data.filstorage.io/tcp/443/https` (fails, due to basic-auth) and `/dns4/dag.w3s.link/tcp/443/https` (open) so the block is likely fetched from `dag.w3s.link`.

src/pages/config.tsx
```diff
@@ -84,10 +103,30 @@ function ConfigPage (): React.JSX.Element | null {
...
   useEffect(() => {
+    hasLocalGateway()
+      .then(async hasLocalGw => {
+        if (hasLocalGw) {
+          // check if local storage has it.
+          const unparsedGwConf = localStorage.getItem(LOCAL_STORAGE_KEYS.config.gateways)
+          const gwConf = unparsedGwConf != null ? JSON.parse(unparsedGwConf) as string[] : defaultGateways
+          if (!gwConf.includes(localGwUrl)) {
+            log(`Adding ${localGwUrl} to gateway list`)
+            gwConf.push(localGwUrl)
```
lidel: nit: prepend to the list, so localhost is prioritized
```suggestion
            gwConf.unshift(localGwUrl)
```

lidel (changes requested):
Did not run it locally yet, but dropping quick feedback / questions in the meantime.

src/lib/local-gateway.ts
2color: Based on https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts 127.0.0.1 should also be Secure Context. 

Is there any particular benefit to using localhost. We don't change the `redirect` config in fetch, so it's follow, and indeed it is followed.  

I don't have a strong opinion, just curious why localhost is preferred.

src/lib/config-db.ts
2color: > If we remove default delegated router, won't we lose opportunistic direct retrieval over HTTP? 

That's correct. 

> Maybe keeping the default as-is is better play for Camp?

In #288 you proposed disabling it by default:

> It should be disabled by default, until we have direct retrieval working

If we leave it there by default, it will continue making a request to the delegated routing endpoint for every block it encounters. 

---
I don't have a strong opinion. Once we add the toggle this will be much easier to disable anyway.

lidel: Yeah, let's keep it, I wrote that before I realized we do have direct retrieval over http if http multiaddr is present.
 See you already restored, so resolving.

2color:
I'm currently blocked by a bug I found, where verifiedFetch does not respect the `allowInsecure` and `allowLocal` config options passed to it.

I have a reproduction here: https://codepen.io/2color/pen/VwOXeMz. But I haven't been able to find out where the root of the problem is. Still digging into this.

lidel (changes requested):
We don't have testing in place for this, so each flow needs to be inspected manually.
I did some testing in vanilla Google Chrome (empty profile each time) and both need a bit more polish.

- **Localhost detection** works only when dancing a specific way:
  - 🟢 Works when visiting the main domain first (`http://test1.localhost:3000/`) and then manually open config  detects and prepends my `localhost:8080` gateway, and uses it just fine
  - 🚧 Does not work when visiting the main domain first with config URL opened (`http://test2.localhost:3000/#/ipfs-sw-config`) 
  - 🚧 Does not work if I visit subdomain first (`http://specs-ipfs-tech.ipns.test3.localhost:3000/` does not prepend localhost)
  - ⭕ The detection does dot work in Brave because they have very strick adblock (Brave Shields) rules: 
    >   
    Due to this, a big chunk of audience won't be able to benefit from this anyway :disappointed: 
- **Disabling delegated routing** works, but does more harm than good, making it not useful:
  - 🚧 Disabling delegated routing breaks DNSLink resolution ([example](https://github.com/ipfs/service-worker-gateway/assets/157609/6b66a431-6b08-4a63-b8e1-e92d4740cd26)), while it should only impact IPNS and not DNSLink

-----

Between localhost detection and adjusting  things in https://github.com/ipshipyard/waterworks-infra/issues/55 the latter improves things in all browsers, so we may want to park this and focus on that + the other quality-of-life improvements from the Camp milestone.

I feel we will get back to this once we simplify codebase after Camp. Right now, too much of a time sink.

src/lib/local-gateway.ts
SgtPooki: I don't think there's any more to it for most usecases.

SgtPooki:
@2color I created https://github.com/ipfs/service-worker-gateway/pull/403 to replace this one. Closing

Issue #21, opened 2024-02-13
PWA with handlers for ipfs:// and ipns://


lidel:
> This is something we could explore when more pressing things are implemented, after we deploy service worker gateway under long-term domain name.

When packaged as [PWA](https://web.dev/learn/pwa/), SW gateway root could double as a very nice light mobile client. If we create a compelling landing page that could double as a mobile app, we would be bringing light IPFS node with built-in verifiable `ipfs://` and `ipns://` URL support to mobile users (android etc) virtually for free.

## Background 

It is possible to create a Progressive Web App (PWA) with a manifest file that declares the app as a handler for specific protocols or URLs. This allows PWAs to provide handler of certain URLs or protocols.

The user experience is pretty good on mobile platforms, especially Android, where [PWA support is built into the operating system](https://support.google.com/chrome/answer/9658361?hl=en&co=GENIE.Platform%3DAndroid).

## Implementation notes and caveats

- Assuming we have subdomain service worker gateway at `example.com`
- Requesting main domain would return HTML with `<link rel="manifest" href="/manifest.json">` and there would be PWA manifest at `example.com/manifest.json`
- `manifest.json` would have `url_handlers` section which defines two handlers that implement [`uri` router from subdomain spec](https://specs.ipfs.tech/http-gateways/subdomain-gateway/#uri-router):
  - `ipfs:` → `/ipfs/?uri=%s` → decode `ipfs://` URL, redirect to proper subdomain
  - `ipns:` → `/ipns/?uri=%s` → decode `ipns://` URL, redirect to proper subdomain
  - note: we may be able to use `#uri` here to avoid sending request to http server, and doing URI decoding and redirect via `window.location.replace`
  - note: `%s` is [percent-encoded URL](https://en.wikipedia.org/wiki/Percent-encoding) with custom scheme like `ipfs://cid/optional/path` making the URL looks like this: `?uri=ipfs%3A%2F%2FQmYizgL5Gn6zZFK8n5thwULT4q12P5yVAT1V4xuicp8226%2Ffavicon-32x32.png`

lidel:
Mostly done in https://github.com/ipfs/service-worker-gateway/pull/310, remaining work here here is to add protocol handlers to manifest:
- https://developer.mozilla.org/en-US/docs/Web/Manifest/protocol_handlers

It should use the same handler as one described in
-  https://github.com/ipfs/service-worker-gateway/issues/214#issuecomment-2102963620

[...]

PR #310, opened 2024-07-01
feat: minimal PWA manifest

lidel:
This PR is the first stab at making it possible to install and run in app mode.

### TODO

- [x] manifest and icon
- [x] land other things from the IPFS Camp milestone
- [x] rebase and tweak any papercuts 
- [ ] add protocol handlers and `?uri=` router (details in #21)

Aims to work towards #21


public/manifest.json
```diff
@@ -0,0 +1,18 @@
+{
+  "short_name": "InBrowser IPFS",
+  "name": "InBrowser IPFS Gateway",
```
lidel, line 3: :thought_balloon: ideas for better name welcome. We need something short, perhaps just go with `inbrowser.link` ?

SgtPooki (commented):
Code looks fine to me but i havent looked at a pwa manifest in a long time. Ill need to deploy local and see if i can get it to work

public/manifest.json
SgtPooki: i think thats about as short as we can go and have it still be semantic

lidel:
@SgtPooki it is super tricky to test on mobile in LAN due to lack of https, but I was able to do it in desktop Google Chrome with `goo.localhost:3000`, it shows small icon on the right side of address bar that allows installing the PWA. 

Something that is missing is [`protocol_handlers`](https://developer.mozilla.org/en-US/docs/Web/Manifest/protocol_handlers), but we need and `?uri=%s` router ([specs](https://specs.ipfs.tech/http-gateways/subdomain-gateway/#uri-router)) for that.

I am thinking I could add router by converting `?uri=%s` into `/ipxs/id/..` path before passing it to `verifiedFetch` [here](https://github.com/ipfs/service-worker-gateway/blob/774e388bd2c4696988121c42ce511c59837348f0/src/sw.ts#L492), but lmk if there is a better way?
It could also be implemented as redirect response upstream in verified fetch, but not sure how useful it is beyond this project.

SgtPooki:
> @SgtPooki it is super tricky to test on mobile in LAN due to lack of https, but I was able to do it in desktop Google Chrome with goo.localhost:3000, it shows small icon on the right side of address bar that allows installing the PWA.

Mac allows you to do simulation of iphone from desktop, i will give this a quick try before i move over to something else

SgtPooki:
> > @SgtPooki it is super tricky to test on mobile in LAN due to lack of https, but I was able to do it in desktop Google Chrome with goo.localhost:3000, it shows small icon on the right side of address bar that allows installing the PWA.
> 
> Mac allows you to do simulation of iphone from desktop, i will give this a quick try before i move over to something else

image

image

image

image

image

SgtPooki:
there are some UX issues and issues with content for specs.ipfs.tech. The simulator history back and forward work, but it's hard to trigger.. in general it seems to work okay!

lidel:
Thanks! Let's ship this, and add protocol handlers at a later time.

PR #311, opened 2024-07-01
chore: add header to most pages

### LINKED ISSUE #309: ux: keep topbar at all times
## Problem

We have different UI on redirect pages, the main landing page, and error pages:

| Landing page (`foo.localhost:3000`) | Redirect / subdomain landing (`http://specs-ipfs-tech.ipns.foo.localhost:3000/`) | Error page (`http://error-test.ipns.foo.localhost:3000/`) | 
| ---- | ---- | ---- |
| Screen Shot 2024-06-28 at 20 34 17 | Screen Shot 2024-06-28 at 20 35 44 | Screen Shot 2024-06-28 at 20 37 16 |

This makes navigation and configuration difficult, especially on mobile / PWA context.

## Proposed improvement

Let's unify all three screens to follow the main landing page.
Keep the top bar with button that allows editing configuration:

image

  
Other improvements:  
- Remove "About" section from the "Redirect / subdomain landing" page (top header will be enough)
  - To keep continuity, we should also stop opening pages in new tab ([remove `target=_blank` from here](https://github.com/ipfs/service-worker-gateway/blob/774e388bd2c4696988121c42ce511c59837348f0/src/components/input-validator.tsx#L83))
- Error page: reuse style from [#272 ](https://github.com/ipfs/service-worker-gateway/pull/283#pullrequestreview-2148710443)

SgtPooki:
## Title

chore: add header to all pages

## Description

This change adds the header to renderable pages. ~~The only page without the header at this point is [the one the service-worker renders when there is an error](https://github.com/ipfs/service-worker-gateway/blob/774e388bd2c4696988121c42ce511c59837348f0/src/sw.ts#L566-L589).~~ 

Fixes https://github.com/ipfs/service-worker-gateway/issues/309

### Config page
image

### Redirect page
image

### Service-worker served Error page

image

In order to render the same header for the service worker error page we will need to compromise somehow.

#### Problems
* We can't really do react/jsx easily from the SW (this would all spike asset size for sw..)
* We can't do dynamic imports (not really) in SW
* build process is a little convoluted and makes building an index file + css for SW a little messy

#### Potential solutions

* (currently coded solution): Copy raw HMTL for header, include `cdn.jsdelivr.net` links for ipfs-css and tachyons..
* The SW does have cache for almost all the other ipfs-sw assets.. so maybe something with the cached assets can be used if we can get the content-hash, or get webpack to fill in some variables for us
* We could just copy over css from dist and import it all from service worker, and add in final rendered HTML for header.. but this will be a maintenance nightmare for the future.
* We could manually copy needed CSS and raw HTML for the header into SW.. but also a potential maintenance pain.

~~- [ ] Need to update sw error page button to anchor tag with appropriate link.. or JS for handling click event on the gear icon.~~

## Change checklist


SgtPooki:
@lidel could use your thoughts on the SW error page when you get a chance

lidel (commented):
@SgtPooki thank you for poking at this an analysis. 
I think bending backwards to get topbar on error pages is not worth it. 

At the end of the day, the minimal UX user has on error page is to be able to "Retry" or "Abort".
With that in mind, perhaps instead of topbar on error pages, could we add two buttons at the top?

-  "Retry" to execute `window.location.reload()`
-  "Go back" to execute `window.history.back()`

SgtPooki:
**Latest**

image

SgtPooki (commented):
self review

lidel (approved):
Thanks! 

While testing, realized we now had two config interlaces, one behind hear, and one in iframe, so pushed small fix (https://github.com/ipfs/service-worker-gateway/pull/311/commits/b9f9ad9ba1f321fa232f451890e22862066bad3b) that hides the duplicated/broken config interface.

PR #315, opened 2024-07-03
feat: config page ux improvements

### LINKED ISSUE #290: ux: make URL lists line-separated textboxes
## Problem

JSON array in input is bad UX, hard to edit, and very error prone:

## Proposed fix

Do what we do in ipfs-companion, have newline-separated textarea that grows a bit based on the content:

> 

>

SgtPooki:
- feat: convert config inputs to textarea
- fix: save config works
- feat: add descriptions to config page inputs
- chore: package-lock.json update

## Title

feat: config page ux improvements

## Description

Updates the config page inputs that were `type=input` to `<textarea>` and adds descriptions to each input. It makes the page more user-friendly and less prone to input errors.

Some style updates were done to resolve a long-standing UX issue with error message placement.

Fixes https://github.com/ipfs/service-worker-gateway/issues/290

## Change checklist

### DIFF
--- a/src/lib/config-db.ts
+++ b/src/lib/config-db.ts
@@ -1,6 +1,6 @@
 import debugLib from 'debug'
 import { GenericIDB, type BaseDbConfig } from './generic-db.js'
-import { LOCAL_STORAGE_KEYS, convertDnsResolverInputToObject, convertUrlInputToArray } from './local-storage.js'
+import { LOCAL_STORAGE_KEYS, convertDnsResolverInputToObject, convertDnsResolverObjectToInput, convertUrlArrayToInput, convertUrlInputToArray } from './local-storage.js'
 import type { ComponentLogger } from '@libp2p/logger'
 
 export interface ConfigDb extends BaseDbConfig {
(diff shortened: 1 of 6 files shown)


src/lib/config-db.ts
```diff
@@ -23,14 +23,14 @@ export async function loadConfigFromLocalStorage (): Promise<void> {
   if (typeof globalThis.localStorage !== 'undefined') {
     await configDb.open()
     const localStorage = globalThis.localStorage
-    const localStorageGatewaysString = localStorage.getItem(LOCAL_STORAGE_KEYS.config.gateways) ?? JSON.stringify(defaultGateways)
-    const localStorageRoutersString = localStorage.getItem(LOCAL_STORAGE_KEYS.config.routers) ?? JSON.stringify(defaultRouters)
-    const localStorageDnsResolvers = localStorage.getItem(LOCAL_STORAGE_KEYS.config.dnsJsonResolvers) ?? JSON.stringify(defaultDnsJsonResolvers)
+    const localStorageGatewaysString = localStorage.getItem(LOCAL_STORAGE_KEYS.config.gateways) ?? convertUrlArrayToInput(defaultGateways)
```
lidel: nit: would it be possible to keep existing behavior (store JSON) and convert to prettier newline delimited representation only in the end users GUI?

Right now, existing users that have preexisting JSON in storage are being broken (screenshot below):

>

lidel (changes requested):
Thanks! Pushed small css tweaks, validation now looks like this:

> 

Small concern about changing storage format (see details below), other than that lgtm.

src/lib/config-db.ts
SgtPooki: totally. let me fiddle with that. i created functions for conversion both ways so that shouldn't be a big deal.

SgtPooki: ok @lidel give it a try with [add703b](https://github.com/ipfs/service-worker-gateway/pull/315/commits/add703b9e46fc13436ef79498835e1d37323cb33)

lidel (approved):
Thank you @SgtPooki, let's ship this :+1:

PR #319, opened 2024-07-05
fix: remove auto-reload config and default to true

SgtPooki:
## Title

fix: remove auto-reload config and default to true

## Description

This PR removes the autoreload config from the UI, and defaults the autoReload internal state to true (config.autoReload isn't used with the changes in this PR)

## Notes & open questions

Autoreload config worked in 7517a620a66aab260c66e0cce2270bc8898a0160 and stopped working in [0c6d6010d9aa6413632039f9f290ffe111422224](https://github.com/ipfs/service-worker-gateway/commit/0c6d6010d9aa6413632039f9f290ffe111422224). (thanks lidel for digging)

## Change checklist


SgtPooki:
e2e tests need updated because default auto-reload changes our expectations in some tests. will get back to this shortly.

lidel:
@SgtPooki thanks! this does the trick. I think it is ok to effectively remove tests of interstitial, it no longer serving purpose of interacting with user beyond indicating load being in progress.

UX idea: with autoreload being the default, I think we don't really need the button either now, perhaps we could replace it with splash/spinner from  https://github.com/ipfs/service-worker-gateway/pull/258

Bit late on my end, and we should not ship on friday/weekend,
so let's try to land this on Monday.

src/pages/config.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useCallback, useEffect, useState } from 'react'
 import Header from '../components/Header.jsx'
 import { Collapsible } from '../components/collapsible.jsx'
 import LocalStorageInput from '../components/local-storage-input.jsx'
-import { LocalStorageToggle } from '../components/local-storage-toggle.jsx'
+// import { LocalStorageToggle } from '../components/local-storage-toggle.jsx'
```
SgtPooki: will remove

src/pages/config.tsx
```diff
@@ -179,15 +179,15 @@ function ConfigPage (): React.JSX.Element | null {
...
+        {/* <div className='f5 ma0 pt3 teal fw4 db'>Interstitials</div>
         <span className="charcoal-muted f6 fw1 db pt1 pb1 lh-copy">Control if visiting a new origin should display interstitial pages or automatically load the content using existing configuration.</span>
         <LocalStorageToggle
           className="e2e-config-page-input e2e-config-page-input-autoreload"
           localStorageKey={LOCAL_STORAGE_KEYS.config.autoReload}
           onLabel='Auto Reload'
           offLabel='Show Config'
           resetKey={resetKey}
-        />
+        /> */}
```
SgtPooki: will remove

src/pages/redirect-page.tsx
```diff
@@ -58,9 +58,9 @@ function RedirectPage ({ showConfigIframe = true }: { showConfigIframe?: boolean
         log.error('redirect-page: error setting config on subdomain', err)
       }
 
-      if (config.autoReload) {
-        setIsAutoReloadEnabled(config.autoReload)
-      }
+      // if (config.autoReload) {
+      //   setIsAutoReloadEnabled(config.autoReload)
+      // }
```
SgtPooki: will remove

SgtPooki (commented):
self review

PR #391, opened 2024-10-18
feat: add p2p retrieval option

2color:
## What's in this PR
- feat: improve the UX of the toggle button
- feat: add config toggle for p2p retrieval
- deps: add helia and bump verified fetch
- feat: load p2p helia
- Attempt to make recursive gateway optional

## TODO

- [x] Check that validation is working correctly
- [x] Add additional toggles detailed below

## Notes & open questions

Following a discussion with @lidel, we should add the following toggles as part of this PR
- "Enable recursive gateways": with toggle you can disable reliance on the configured gateways
- Transports:
  - Enable WSS
  - Enable WebTransport
  - Enable HTTP Gateways Providers

As for the actual implementation, investigate whether this to do handle these toggles with IPIP-484 filtering (which would only work with delegated routing) **or** adjusting the libp2p instance configuration.

## Change checklist


src/lib/config-db.ts
```diff
@@ -96,9 +107,9 @@ export async function getConfig (logger: ComponentLogger): Promise<ConfigDb> {
     log('error loading config from db', err)
   }
```
2color: We probably just want an additional toggle that enables and disables recursive gateways.

src/components/local-storage-toggle.tsx
```diff
@@ -28,11 +37,40 @@ export const LocalStorageToggle: React.FC<LocalStorageToggleProps> = ({ resetKey
...
+        className={`relative dib h2 w3 flex-shrink-0 pointer br4 ${
+          isChecked ? 'bg-green' : 'bg-dark-gray'
+        } transition-all duration-200 ease-in-out`}
+      >
+        <span
+          className={`absolute top-0 left-0 dib h2 w2 br-100 bg-white shadow-1 ${
+            isChecked ? 'translate-x-100' : 'translate-x-0'
+          } transition-transform duration-200 ease-in-out`}
+        />
       </label>
```
SgtPooki: I think we should reduce the size of this toggle, it looks a little too large right now?

changing the size from `h2 w2` to `h1 w1` (for the span) and from `h2 w3` to `h1 w2` (for the label) looks a little more appealing to me.. what do you think?

image

SgtPooki: Also, do we need to change the styles of these toggles? we had a review of the toggle UI a while back

src/components/local-storage-toggle.tsx
```diff
@@ -28,11 +37,40 @@ export const LocalStorageToggle: React.FC<LocalStorageToggleProps> = ({ resetKey
...
+
     </div>
   )
 }
+
+function getLocalStorageValue (localStorageKey: string, defaultValue: boolean): boolean {
+  const savedValue = localStorage.getItem(localStorageKey)
+  if (savedValue == null) {
+    localStorage.setItem(localStorageKey, String(defaultValue))
```
SgtPooki: I think setting defaults here should be avoided. we should set defaults in config-db.ts as all the others are done

src/lib/get-helia.ts
```diff
@@ -0,0 +1,122 @@
...
+import type { ComponentLogger } from '@libp2p/logger'
+
+export async function getVerifiedFetch (config: ConfigDb, logger: ComponentLogger): Promise<VerifiedFetch> {
+  const log = logger.forComponent('get-verified-fetch')
+  log(`config-debug: got config for sw location ${self.location.origin}`, config)
+
+  if (!config.enableRecursiveGateways && !config.enableGatewayProviders && !config.enableWss && !config.enableWebTransport) {
+    // crude validation
+    throw new Error('Config is invalid. At least one of the following must be enabled: recursive gateways, gateway providers, wss, or webtransport.')
+  }
```
SgtPooki: Can we do this validation when config save button is enabled/clicked instead of here in `getVerifiedFetch`? I believe we have logic for that already

src/pages/config.tsx
```diff
@@ -136,34 +137,67 @@ function ConfigPage (): React.JSX.Element | null {
...
+        />
+        <LocalStorageToggle
+          className="e2e-config-page-input"
+          label="Enable WebTransport Providers"
+          description="Use WebTransport providers returned from delegated routers for direct retrieval."
+          defaultValue={defaultEnableWebTransport}
+          localStorageKey={LOCAL_STORAGE_KEYS.config.enableWebTransport}
+          resetKey={resetKey}
+        />
+        {/* TODO: we should only accept a single router URL */}
```
SgtPooki: Is this true? aren't there cases where we may want to use multiple routers?

src/lib/get-helia.ts
```diff
@@ -0,0 +1,122 @@
+import { noise } from '@chainsafe/libp2p-noise'
```
SgtPooki: we should rename this file to `get-verified-fetch.ts`

src/pages/config.tsx
```diff
@@ -136,34 +137,67 @@ function ConfigPage (): React.JSX.Element | null {
...
+        localStorageKey={LOCAL_STORAGE_KEYS.config.enableRecursiveGateways}
+        resetKey={resetKey}
+      />
         <LocalStorageInput
           className="e2e-config-page-input e2e-config-page-input-gateways"
-          description="A newline delimited list of trustless gateway URLs."
+          description="A newline delimited list of recursive trustless gateway URLs."
           localStorageKey={LOCAL_STORAGE_KEYS.config.gateways}
-          label='Gateways'
+          label='Recursive Gateways'
```
SgtPooki: A gateway could be recursive without being trustless, yes? But a recursive, non-trustless, gateway wouldn't work for verified-fetch calls

```suggestion
          label='Trustless Gateways'
```

src/pages/config.tsx
```diff
@@ -136,34 +137,67 @@ function ConfigPage (): React.JSX.Element | null {
...
     <main className='e2e-config-page pa4-l bg-snow mw7 center pa4'>
       <Collapsible collapsedLabel="View config" expandedLabel='Hide config' collapsed={isLoadedInIframe}>
+      <LocalStorageToggle
+        className="e2e-config-page-input"
+        label="Enable Recursive Gateways"
+        description="Use recursive gateways configured below for retrieval of content."
+        defaultValue={defaultEnableRecursiveGateways}
+        localStorageKey={LOCAL_STORAGE_KEYS.config.enableRecursiveGateways}
+        resetKey={resetKey}
+      />
```
SgtPooki:
```suggestion
	      <LocalStorageToggle
	        className="e2e-config-page-input"
	        label="Enable Trustless Gateways"
	        description="Use recursive trustless gateways configured below for retrieval of content."
	        defaultValue={defaultEnableRecursiveGateways}
	        localStorageKey={LOCAL_STORAGE_KEYS.config.enableRecursiveGateways}
	        resetKey={resetKey}
	      />
```

SgtPooki (commented):
a few minor comments and questions and suggestions. will run locally and look into it

src/components/local-storage-toggle.tsx
2color: The old toggle was constantly confusing me.
I don't think thing having text helps, because you can still read the text of the disabled state:


Screenshot 2024-10-22 at 6 01 47 PM
Screenshot 2024-10-22 at 6 02 45 PM

src/components/local-storage-toggle.tsx
2color: Yeah, I can't remember why exactly that was necessary, but I agree. I'll look into this.

src/pages/config.tsx
2color: Can you think of such a case? 

It seems unnecessary optionality.

Moreover if libp2p is enabled, we rely on libp2p routing exclusively, which doesn't support multiple delegated routers.

src/lib/get-helia.ts
2color: Yeah good point. That was a TODO.We don't have global validation implemented. But I'll take care of that.

src/pages/config.tsx
2color: It's mostly to make a distinction from **non-recursive** gateway providers.

Recursive is here to indicate that the gateway is handling routing/retrieval for you.

src/pages/config.tsx
2color: Just to clarify, this is only concerned with recursive gateways. You can have this enabled but still allow (trustless) gateway providers.

src/components/local-storage-toggle.tsx
2color: Binary toggles with a label that describes their behaviour in binary terms are more easily understandable.

As for the size. I think it's a good size currently, more or less the same as the old toggle and the save button.

src/lib/get-helia.ts
SgtPooki: I thought we had validation before save was clicked.. the button was disabled if inputs were invalid.. we can save until later if that's the case.

src/components/local-storage-toggle.tsx
SgtPooki: ya you're right. I do like the new toggle better. what do you think about making it slightly smaller?

src/pages/config.tsx
SgtPooki: I guess that's true, but are all trustless gateways recursive? maybe we can just add that in the description?

src/pages/config.tsx
SgtPooki: There are two things in this suggestion that may not be clear:

1. indentation
2. the change of the label.

indentation should be fixed for sure.

as for the label, how do we have trustless gateway providers if they're named "recursive gateways"?

if we call a trustless gateway for content, we don't control whether they are recursing to retrieve that content or not.. right? I'm slightly confused by the distinction between recursive and trustless gateways with these updates.

src/pages/config.tsx
SgtPooki: I was thinking of the fact that I might want to use a public delegated router, and a private one.. sure, they could set up their own gateway, but we should be able to call varying authorities (routers) to figure out who has content

src/components/local-storage-toggle.tsx
2color: Yeah I'll try making it smaller

src/pages/config.tsx
2color: No. The classic example is w3link which only returns blocks it has. Almost all the gateways coming from the ipni are non recursive and essential just serve cached blocks over a trustless gateway

src/pages/config.tsx
2color: That's fair, maybe we need to work more on the copy here. 


To start off: all gateways used by verified fetch are trustless. 

Recursive gateways are more expensive to run because they handle retrieval of blocks they don't have. 

Non recursive gateways are typically providers announcing to the IPNI, who serve blocks over http using the trustless gateway. It's typically a cheaper alternative to making the blocks available over bitswap. It's pretty common for Filecoin providers. (The reason that some opt for bitswap is because kubo doesn't support http retrievals yet)

src/components/local-storage-toggle.tsx
2color: >changing the size from h2 w2 to h1 w1 (for the span) and from h2 w3 to h1 w2 (for the label) looks a little more appealing to me.. what do you think?


Seemed a little too small to my taste, but they're about the same size as the built in macOS toggles, so I'll defer to Apple's design rigour  😅
Screenshot 2024-10-23 at 4 28 01 PM

src/components/local-storage-toggle.tsx
2color: Fixed while pairing

src/lib/get-helia.ts
2color: Resolving this in favour of https://github.com/ipfs/service-worker-gateway/issues/399

src/pages/config.tsx
lidel: My mental model is this:
- Recursive gateways are the "last resort", a fallback. 
  - Ideally, we would have ability to delay hitting them, and try direct retrieval first. 
  - SW comes with a list of those fallback gateways. In our case that will only be `https://trustless-gateway.link`.
- SW may discover other HTTP gateways via routing system 
  - e.g. `/dns4/dag.w3s.link/tcp/443/https` [here](https://delegated-ipfs.dev/routing/v1/providers/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi)
  - if we discover HTTP gateway in relation to a specific CID
    - it should NOT be added to the list of recursive (fallback) gateways
    - should only used as provider of that specific CID (or DAG, if we keep session info around).
- UI-wise, we want to 
  - use recursive gateways only as a last resort fallback, when direct retrieval from actual provider fails or is not possible, 
  - and have toggle to disable that fallback, so we can see when direct retrieval fails.
  - avoid generic terms like "trustless gateway", just like we would not use "TCP socket address" without any context about the function of the socket :)

To avoid confusion, perhaps we could call things what they are, in order: 
- First, have section `Direct Retrieval` with toggles of supported retrieval options:
  - `Delegated Routing` list of endpoints, with `Use delegated routers to find content providers` toggle.
  - `Support WebSockets`
  - `Support WebTransport`
  - `Support HTTPS` that controls if we act on gateways discovered via routing system
- Then, at bottom have `Fallback Retrieval` section
  - `Recursive HTTPS Gateways` list
    - And a toggle next to the list to `Use as fallback if direct/p2p retrieval is not possible`
  - By the fact its at the end, it conveys it is a last resort fallback, and not the golden path

src/pages/config.tsx
2color: To be resolved later in https://github.com/ipfs/service-worker-gateway/issues/400

src/pages/config.tsx
SgtPooki: config sections done in [`7e7ef6f` (#391)](https://github.com/ipfs/service-worker-gateway/pull/391/commits/7e7ef6f8db4098fa4d6144d7e34ce88a1f0f53d2)

src/components/input-section.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+
+export const InputSection: React.FC<{ label: string, children: React.ReactNode[] }> = ({ label, children }) => {
+  return (
+    <div className="bl bw2 br2 b--teal-muted pl2 mb3">
```
2color:
```suggestion
    <div className="bl bw2 br2 b--teal-muted pl3 mb3">
```
Improves readability 

Before:
Screenshot 2024-10-24 at 1 51 14 PM
After 
Screenshot 2024-10-24 at 1 51 25 PM

src/pages/config.tsx
2color: @lidel regarding using recursive gateways as a fallback, I opened a new issue in the Helia repo to discuss this, because it's currently not easily done natively in Helia.

See https://github.com/ipfs/helia/issues/665

2color:
@SgtPooki I think this is good to merge now. 

I've re-enabled support for multiple routers.

SgtPooki (approved):

PR #392, opened 2024-10-21
feat: build with esbuild, apply source maps

SgtPooki:
## Title

feat: build with esbuild

## Description

* Removes webpack
* builds with esbuild, adds source maps
* ~~NOTE: Code-splitting is not implemented at this time. SW imports break with code-splitting because we're missing magic that webpack was doing to convert `import`s to `importScripts`~~
    * If you have an old SW installed, the new sw that supports import statements will not work and will tell you "cannot use import statements". You must uninstall the old SW to get things working

## Notes & open questions

The latest changes here are deployed to https://sw.sgtpooki.com

### Potential TODO:

- [ ] We may want to remove some fonts from the `dist/` folder, but I don't believe they're being inlined.
- [ ] Investigate firefox issue when `splitting: true`

### dist size

```
╰─ ✔ ❯ du -h dist                                                                                            4.21   26.1G   65%   100%  ─╯
14M	dist

╰─ ✔ ❯ du -h dist/*.js                                                                                       4.21   26.2G   65%   100%  ─╯
196K	dist/ipfs-sw-index-J6NQTQUO.js
504K	dist/ipfs-sw-sw.js

╰─ ✔ ❯ du -h dist/*.css                                                                                      4.15   26.2G   65%   100%  ─╯
92K	dist/ipfs-sw-index-2QCJAGKE.css
```

## Change checklist


test/node.ts
```diff
@@ -43,15 +43,15 @@ describe(`verify-dist at ${distRoot}`, () => {
...
       filePath: 'index.html',
       content: [
-        /ipfs-sw-styles-.*\.css/,
-        /ipfs-sw-styles-.*\.js/,
-        /ipfs-sw-main-.*\.js/,
-        /ipfs-sw-vendor-react-.*\.js/,
+        /ipfs-sw-index-.*\.css/,
+        // /ipfs-sw-styles-.*\.js/,
+        // /ipfs-sw-main-.*\.js/,
+        // /ipfs-sw-vendor-react-.*\.js/,
```
SgtPooki:
```suggestion
```

SgtPooki (commented):
self review

SgtPooki:
@2color This should make debugging problems with https://github.com/ipfs/service-worker-gateway/pull/391 more easy

src/service-worker-utils.ts
```diff
@@ -5,7 +5,9 @@ import { uiLogger } from './lib/logger.js'
  */
 export async function registerServiceWorker (): Promise<ServiceWorkerRegistration> {
   const log = uiLogger.forComponent('register-service-worker')
-  const swRegistration = await navigator.serviceWorker.register(new URL(/* webpackChunkName: "sw" */'sw.js', import.meta.url))
+  const swRegistration = await navigator.serviceWorker.register(new URL('ipfs-sw-sw.js', import.meta.url), {
+    type: 'module'
```
2color: We might have trouble with Firefox.
https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorker#browser_compatibility

Have you tried setting the esbuild bundle format to `iife`? https://esbuild.github.io/api/#format-iife

2color: Upon a closer look, I think this may be the reason you had problems with Firefox when `splitting: true`. 

[Since esbuild only supports `esm` output format when splitting is set to `true`](https://esbuild.github.io/api/#splitting), we should remove the `{ type: 'module' }`

2color:
I looked into removing hashes from outputs altogetherin https://github.com/ipfs/service-worker-gateway/pull/393, but then realised it's probably not worth it, as it will open the window to stale cache responses for sw assets from the service worker.

2color:
>  We may want to remove some fonts from the dist/ folder, but I don't believe they're being inlined.

Not only are they not inlined, we override the default styles so that they are never actually used. 

https://github.com/ipfs/service-worker-gateway/blob/2710f1ef3c942d2298e3e57f29ab686887bec3f6/src/pages/default-page-styles.css#L4-L5

2color (approved):
I tested this locally a fair bit. LGTM

src/service-worker-utils.ts
2color: Fixed in https://github.com/ipfs/service-worker-gateway/pull/392/commits/4225e61aa11daa4eaa7c65a433c014571901a686

public/index.html
```diff
@@ -20,7 +20,7 @@
     <meta charset="UTF-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title><%= htmlWebpackPlugin.options.title %> | <%= htmlWebpackPlugin.options.version %></title>
+    <title>IPFS Service Worker Gateway | <%= GIT_VERSION %></title>
```
SgtPooki, line 23: @2color thanks for this, i was planning to hit this today

PR #394, opened 2024-10-22
fix: remove fonts from dist folder

SgtPooki:
## Title

fix: remove fonts from dist folder

## Description

Removes ipfs-css fonts from the dist folder. We are not using them and they are large.

see https://github.com/ipfs/service-worker-gateway/pull/392#issuecomment-2429029055

```shell
╰─ ✔ ❯ ls dist
_redirects    index.html		      ipfs-sw-index-IEJROQ5T.js.map   ipfs-sw-ipfs-cube-O6GHTSYO.svg  ipfs-sw-sw.js.map
favicon.ico   ipfs-sw-gear-icon-THUWFABE.svg  ipfs-sw-index-LKQONJJF.css      ipfs-sw-ipfs-logo-CWKV2PO5.svg  manifest.json
icon-512.png  ipfs-sw-index-IEJROQ5T.js       ipfs-sw-index-LKQONJJF.css.map  ipfs-sw-sw.js

╰─ ✔ ❯ du -h dist
4.5M	dist
```

## Change checklist

### DIFF
--- a/build.js
+++ b/build.js
@@ -51,6 +51,7 @@ function gitRevision () {
     return `no-git-dirty@${new Date().getTime().toString()}`
   }
 }
+
 /**
  * Inject the dist/index.js and dist/index.css into the dist/index.html file
  *
@@ -130,6 +131,21 @@ const modifyBuiltFiles = {
 }
 
 /**
+ * @param {string[]} extensions - The extension of the imported files to exclude. Must match the fill ending path in the import(js) or url(css) statement.
+ * @returns {esbuild.Plugin}
+ */
+const excludeFilesPlugin = (extensions) => ({
+  name: 'exclude-files',
+  setup (build) {
+    build.onResolve({ filter: /.*/ }, async (args) => {
+      if (extensions.some(ext => args.path.endsWith(ext))) {
+        return { path: args.path, namespace: 'exclude', external: true }
+      }
+    })
+  }
+})
+
+/**
  * @type {esbuild.BuildOptions}
  */
 export const buildOptions = {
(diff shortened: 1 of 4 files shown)


SgtPooki (commented):
self review

2color (approved):

build.js
```diff
@@ -129,6 +130,21 @@ const modifyBuiltFiles = {
...
+/**
+ * @param {string[]} extensions - The extension of the imported files to exclude. Must match the fill ending path in the import(js) or url(css) statement.
+ * @returns {esbuild.Plugin}
+ */
+const excludeFilesPlugin = (extensions) => ({
+  name: 'exclude-files',
+  setup (build) {
+    build.onResolve({ filter: /.*/ }, async (args) => {
+      if (extensions.some(ext => args.path.endsWith(ext))) {
+        return { path: args.path, namespace: 'exclude', external: true }
```
2color, line 142: What's the `namespace` property?

SgtPooki: i don't believe it's needed here, but https://esbuild.github.io/plugins/#namespaces talks more about it. 

Essentially, everything is `file` namespace by default.. but you can change the namespace so different plugins can operate on things. We only have custom plugins here, and none of ours are using that namespace, so we could remove it if we want, but, for example, we could filter on namespaces and then simply have a plugin that adds `external: true` for anything in the `exclude` namespace.. see https://esbuild.github.io/plugins/#on-resolve-options

SgtPooki: it allows for modularity

PR #395, opened 2024-10-22
fix: service worker logs are visible again

SgtPooki:
## Title

fix: service worker logs are visible again

## Description

@libp2p/logger has some issue with rendering inside of a service worker. I forgot about this bug when we switched to @libp2p/logger and now just found it again while looking at https://github.com/ipfs/service-worker-gateway/pull/391

## Change checklist

### DIFF
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -8,9 +8,6 @@ const createlibp2pLogger = (prefix: string): Logger => {
   const mainLogger = debug(prefix)
   return Object.assign(debug(prefix), {
     error: mainLogger.extend('error'),
-    // warn: mainLogger.extend('warn'),
-    // info: mainLogger.extend('info'),
-    // debug: mainLogger.extend('debug'),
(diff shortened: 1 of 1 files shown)


src/lib/logger.ts
```diff
@@ -1,5 +1,26 @@
...
+/**
+ * Hack to get around @libp2p/logger not working inside a service worker.
+ */
+const createlibp2pLogger = (prefix: string): Logger => {
+  const mainLogger = debug(prefix)
+  return Object.assign(debug(prefix), {
+    error: mainLogger.extend('error'),
+    // warn: mainLogger.extend('warn'),
+    // info: mainLogger.extend('info'),
+    // debug: mainLogger.extend('debug'),
```
SgtPooki:
```suggestion
```

SgtPooki (commented):
self review

SgtPooki:
@2color this fixes SW logging

PR #397, opened 2024-10-22
feat: re-enable code splitting

SgtPooki:
- **feat: add code splitting back**
- **chore: debugging**
- **chore: debugging**
- **chore: debugging**
- **fix: firefox properly replaces old service worker**

To debug this, follow this process:

```shell
npm run clean && npm run start
# load the page

# if you need to reload the old SW...
git checkout main
npm run clean && npm run start
# load the page

git checkout fix/splitting-sw
npm run clean && npm run start
# load the page

# repeat as needed
```


SgtPooki:
not sure why chrome is failing here.. we can address when code-splitting becomes a priority again

2color:
Do you have any rough numbers on the size of bundle before/after? This can probably help us evaluate the benefit we get out of code splitting.

PR #398, opened 2024-10-23
fix: use enable from @libp2p/logger

SgtPooki:
## Title

fix: use enable from @libp2p/logger

## Description

@libp2p/logger was switched from `debug` to `weald`. We need to use the `enable` function from `@libp2p/logger` to enable logging.

image

Related https://github.com/ipfs/service-worker-gateway/pull/395
Related https://github.com/libp2p/js-libp2p/issues/2778#issuecomment-2431174902

## Change checklist


SgtPooki (commented):
self review

2color (approved):

PR #403, opened 2024-10-24
feat: probe for local gateway and add to config

### LINKED ISSUE #288: ux: add configuration toggle for delegated routing

### LINKED ISSUE #299: ux: detect local trustless gw and default to it

SgtPooki:
## Title

feat: probe for local gateway and add to config

## Description

Replaces https://github.com/ipfs/service-worker-gateway/pull/303
Fixes https://github.com/ipfs/service-worker-gateway/pull/288
Fixes https://github.com/ipfs/service-worker-gateway/pull/299

## Notes & open questions

~~Not tested. Only created because merge conflicts on https://github.com/ipfs/service-worker-gateway/pull/303 and git saying `no related history`~~

This has been tested, and the local gateway check is working properly. I am thinking that we need to add a UI toggle to disable adding it, but we might need to refactor the local-storage inputs and lift up the accessing of those inputs.. or handle the adding of the localStorage gateway when we save

## Change checklist

### DIFF
--- a/src/components/local-storage-toggle.tsx
+++ b/src/components/local-storage-toggle.tsx
@@ -14,6 +14,8 @@ interface LocalStorageToggleProps {
   className?: string
   onClick?: React.MouseEventHandler<HTMLDivElement>
   resetKey: number
+  disabled?: boolean
+  handleChange?(toggleValue: boolean): void
 }
 
 export const LocalStorageToggle: React.FC<LocalStorageToggleProps> = ({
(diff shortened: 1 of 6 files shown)


SgtPooki:
this may need solved later, closing.

PR #405, opened 2024-10-25
fix: support discovery of /tls/../ws|http  providers

lidel:
Opening this to start discussion, this may not be the right fix, see "Solution (B)" below.

cc @SgtPooki @2color for feedback / ideas 

## Problem description

We already support DNS-dependent `/wss` like `/dns4/sv15.bootstrap.libp2p.io/tcp/443/wss/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJ`, but we are filtering out `/ip.../tls/sni/../ws` where DNS resolution is not required. Those will be all p2p-forge peers once we deploy it in Kubo in a few releases.

Example: `/ip4/147.75.87.65/tcp/4002/tls/sni/147-75-87-65.k51qzi5uqu5dhb03btheentd9nmkaygwm4q3tfrm596s5thf1u87bc7gq0m2ia.libp2p.direct/ws/p2p/12D3KooWCrBiagtZMzpZePCr1tfBbrZTh4BRQf7JurRqNMRi8YHF`

### Solution A (this PR)

This PR ensures we are not filtering them out. Compare:

- `curl -H 'accept: application/x-ndjson' 'https://delegated-ipfs.dev/routing/v1/providers/bafybeihn2f7lhumh4grizksi2fl233cyszqadkn424ptjajfenykpsaiw4?filter-addrs=wss'`

with

- `curl -H 'accept: application/x-ndjson' 'https://delegated-ipfs.dev/routing/v1/providers/bafybeihn2f7lhumh4grizksi2fl233cyszqadkn424ptjajfenykpsaiw4?filter-addrs=ws,wss'`

or 

- `curl -H 'accept: application/x-ndjson' 'https://delegated-ipfs.dev/routing/v1/providers/bafybeihn2f7lhumh4grizksi2fl233cyszqadkn424ptjajfenykpsaiw4?filter-addrs=wss,tls'`

This comes with assumption that js-libp2p is smart enough to only dial `/ws` that have `/tls`, and skip  / log error when dialing unencrypted ones, similar to trustless gateways being announced with `/tls/.../http` 
If not, that has to be fixed upstream as well.

Feedback welcome. Alternative approach is (B) below.

### Solution (B) 

Augment `/routing/v1` implementation in boxo server (and someguy) and specs and make it clear that some filters have special meaning:
- `wss` should also include `/tls/../ws`
- `https` should also include `/tls/../http`

This is kinda easy, we can deploy this fix next week.

This feels more future-proof. I'm leaning towards (B) because it shifts responsibility of "doing the right thing" to /routing/v1 endpoint, and ensures even poorly written clients (other than ones we maintain) that dont know about `/tls/..` variants get these results. But lmk your thoughts.

### DIFF
--- a/src/lib/get-verified-fetch.ts
+++ b/src/lib/get-verified-fetch.ts
@@ -68,8 +68,11 @@ export async function getVerifiedFetch (config: ConfigDb, logger: ComponentLogge
   } else {
     config.routers.forEach((router) => {
       routers.push(delegatedHTTPRouting(router, {
+        // NOTE: in practice 'transport-ipfs-gateway-http' exists only in IPNI results, we won't have any DHT results like this unless..
+        // TODO: someguy starts doing active probing to identify peers which have funcitonal HTTP gateway
         filterProtocols: ['transport-ipfs-gateway-http'],
-        filterAddrs: ['https']
+        // Include both /https && /tls/../http
+        filterAddrs: ['https', 'tls']
       }))
     })
 
@@ -83,14 +86,13 @@ export async function getVerifiedFetch (config: ConfigDb, logger: ComponentLogge
   return createVerifiedFetch(helia, { contentTypeParser })
 }
 
-type Libp2pDefaultsOptions = Pick<ConfigDb, 'routers' | 'enableWss' | 'enableWebTransport'>
+type Libp2pDefaultsOptions = Pick<ConfigDb, 'routers' | 'enableWss' | 'enableWebTransport' | 'enableGatewayProviders'>
 
 export async function libp2pDefaults (config: Libp2pDefaultsOptions): Promise<Libp2pOptions> {
   const agentVersion = `@helia/verified-fetch ${libp2pInfo.name}/${libp2pInfo.version} UserAgent=${globalThis.navigator.userAgent}`
   const privateKey = await generateKeyPair('Ed25519')
 
-  const filterAddrs = ['https']
-
+  const filterAddrs = [] as string[]
   const transports: Array<(components: any) => any> = []
 
   if (config.enableWss) {
@@ -101,6 +103,10 @@ export async function libp2pDefaults (config: Libp2pDefaultsOptions): Promise<Li
   if (config.enableWebTransport) {
     transports.push(webTransport())
     filterAddrs.push('webtransport')
+  }
+  if (config.enableGatewayProviders) {
+    filterAddrs.push('https') // /dns/example.com/tcp/443/https
+    filterAddrs.push('tls') // /ip4/A.B.C.D/tcp/4002/tls/sni/example.com/http
   }
 
   const libp2pOptions = {
--- a/src/lib/get-verified-fetch.ts
+++ b/src/lib/get-verified-fetch.ts
@@ -69,7 +69,7 @@ export async function getVerifiedFetch (config: ConfigDb, logger: ComponentLogge
     config.routers.forEach((router) => {
       routers.push(delegatedHTTPRouting(router, {
         // NOTE: in practice 'transport-ipfs-gateway-http' exists only in IPNI results, we won't have any DHT results like this unless..
-        // TODO: someguy starts doing active probing to identify peers which have funcitonal HTTP gateway
+        // TODO: someguy starts doing active probing (https://github.com/ipfs/someguy/issues/53) to identify peers which have functional HTTP gateway
         filterProtocols: ['transport-ipfs-gateway-http'],
         // Include both /https && /tls/../http
         filterAddrs: ['https', 'tls']
--- a/src/lib/get-verified-fetch.ts
+++ b/src/lib/get-verified-fetch.ts
@@ -95,7 +95,8 @@ export async function libp2pDefaults (config: Libp2pDefaultsOptions): Promise<Li
 
   if (config.enableWss) {
     transports.push(webSockets())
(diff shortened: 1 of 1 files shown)


src/lib/get-verified-fetch.ts
```diff
@@ -95,7 +95,8 @@ export async function libp2pDefaults (config: Libp2pDefaultsOptions): Promise<Li
 
   if (config.enableWss) {
     transports.push(webSockets())
-    filterAddrs.push('wss')
+    filterAddrs.push('wss') // /dns4/sv15.bootstrap.libp2p.io/tcp/443/wss/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJ
+    filterAddrs.push('ws') // /ip4/A.B.C.D/tcp/4002/tls/sni/A-B-C-D.peerid.libp2p.direct/ws/p2p/peerid
```
2color:
```suggestion
    filterAddrs.push('tls') // /ip4/A.B.C.D/tcp/4002/tls/sni/A-B-C-D.peerid.libp2p.direct/ws/p2p/peerid
```

2color:
I added a suggestion, but it seems like we could filter these multiaddrs with the `tls` filter. 

The only problem I see with adding `ws` is potentially getting non secure multiaddrs.

----


I'm reluctant with the proposed Solution B, because it changes the assumptions of IPIP-484 (treating filters as opaque strings) and introduces "magic".

src/lib/get-verified-fetch.ts
lidel: I guess this works for this project (we get both `/ws` and `/http` that work in secure context)

lidel: But wait, this way we co-mingle `/tls/../ws` and `/tls/../http`. I think it is ok, because we have decidated flag that disables each transport, so no harm in having extra  results that are not actionable. Just need to make sure we do the same `/tls` trick for HTTP ones.

lidel: Pushed https://github.com/ipfs/service-worker-gateway/pull/405/commits/36d006638297836a32d6e1f4d95e1b9030abaf7a to ensure we also include `/tls` for HTTP ones.

2color: Sounds good.

2color (approved):
LGTM

2color:
We should probably apply these to https://github.com/ipfs/helia/blob/main/packages/helia/src/utils/libp2p-defaults.browser.ts#L66-L68

lidel:
Fair enough – i've refactored  this PR to do filtering on the client, by matching `/tls` in addition to `/wss` and `/https`.

This will comingle both `/http` and `/ws` results but I think it is acceptable overhead. 
The client won't act on those extra `/ws` results  if only HTTP is enabled, and vice-versa. 

If this looks ok, we should ship it as a patch release to allow inbrowser.link fetch wikipedia and other websites from our collab cluster over bitswap+websockets:

> `curl -H 'accept: application/x-ndjson' 'https://delegated-ipfs.dev/routing/v1/providers/bafybeihn2f7lhumh4grizksi2fl233cyszqadkn424ptjajfenykpsaiw4?filter-addrs=wss,tls'`

lidel:
@2color yes, if we don't fixup at server, helia defaults need to be updated to include `tls` – mind opening a PR there?

```diff
- filterAddrs: ['https', 'webtransport', 'webrtc', 'webrtc-direct', 'wss']
+ filterAddrs: ['https', 'webtransport', 'webrtc', 'webrtc-direct', 'wss', 'tls']
```

SgtPooki:
Does enabling tls filter also return tcp tls multiaddr? That should get filtered if tcp is not enabled by js-libp2p but we might want to add `!tcp`?

lidel:
@SgtPooki  hm.. given protocols we have today, virtually every TLS multiaddr for WSS or HTTPS will be on top of TCP (I don't count UDP HTTP/3 as its not used by anyone in ecosystem today). 
If we filter out TCP, we will not have anything left for WS and HTTP to work with. 
Not every TCP is TLS, so with `filter-addrs=tls` we only get browser-safe subset of TCP.

src/lib/get-verified-fetch.ts
```diff
@@ -68,8 +68,11 @@ export async function getVerifiedFetch (config: ConfigDb, logger: ComponentLogge
   } else {
     config.routers.forEach((router) => {
       routers.push(delegatedHTTPRouting(router, {
+        // NOTE: in practice 'transport-ipfs-gateway-http' exists only in IPNI results, we won't have any DHT results like this unless..
+        // TODO: someguy starts doing active probing to identify peers which have funcitonal HTTP gateway
```
SgtPooki: we should probably link to a someguy issue

SgtPooki (approved):

src/lib/get-verified-fetch.ts
lidel: Added `P1: confirm when /tls/../http and /https is a trustless gateway, and add transport-ipfs-gateway-http to Protocols list in respective peer` to https://github.com/ipfs/someguy/issues/53 and referenced here:

```suggestion
        // TODO: someguy starts doing active probing (https://github.com/ipfs/someguy/issues/53) to identify peers which have functional HTTP gateway
```

PR #421, opened 2024-11-05
feat: add blake3 hasher

2color:
## Description

- Add blake3 hasher by default

## Change checklist


SgtPooki (approved):
confirmed that this does work for http://127.0.0.1:8345/ipns/libstc-cc, even though the loaded website doesn't work at that point.. at least the hashing is happening properly (maybe? is the content being borked?)

PR #459, opened 2024-11-14
test: config propagation to subdomains

SgtPooki:


test-e2e/config-loading.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    routers: [process.env.KUBO_GATEWAY as string, 'http://example.com/routing/v1'],
+    dnsJsonResolvers: {
+      '.': 'example.com/dns-query'
+    },
+    debug: 'testDebug',
+    enableWss: false,
+    enableWebTransport: true,
+    enableRecursiveGateways: false,
+    enableGatewayProviders: false
+  }
```
SgtPooki, line 18: I attempted to set non-default values here just so we can be sure it's overriding what would be set without any changes

test-e2e/config-loading.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  test('setting the config actually works', async ({ page, baseURL }) => {
+    await page.goto(baseURL, { waitUntil: 'networkidle' })
+    await waitForServiceWorker(page)
+
+    await setConfig({ page, config: testConfig })
+    expect(await getConfig({ page })).toEqual(testConfig)
```
SgtPooki, line 24: this is really just validating our `setConfig` function on the root domain

SgtPooki (commented):
self review

PR #487, opened 2024-11-22
chore: use groups for dependabot.yml

SgtPooki:
related to https://github.com/ipfs/aegir/pull/1682

## Title

chore: use groups for dependabot.yml

## Description

Groups dependabot updates to:

1. reduce dependabot spam
2. Include related deps that should be updated together

## Change checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,9 @@
 version: 2
 updates:
 - package-ecosystem: npm
-  directory: "/"
+  directories:
+    - "/"
+    - "/packages/*"
   schedule:
     interval: daily
     time: "10:00"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,17 +1,27 @@
 version: 2
 updates:
 - package-ecosystem: npm
-  directory: "/"
+  directories:
+    - "/"
+    - "/packages/*"
```
SgtPooki, line 6: this is not a mono-repo, but I just copied the config over from https://github.com/ipfs/helia-verified-fetch/blob/3a299029ca4c276c49b9b8feb6a5e33e8cc8aba0/.github/dependabot.yml

SgtPooki (commented):
self review

PR #499, opened 2024-12-05
ci: create a deploy workflow which force pushes to staging/production

galargh:
## Title

ci: create a deploy workflow which force pushes to staging/production

## Description

This PR implements the new deployment flow based on the discussions with the team. 

It replace the release please PR flow with a tandem of workflows that force push a tag present on main to the staging/production branch.

## Notes & open questions

It introduces a number of protections:
1. It checks whether a tag to deploy is present in the history of main.
2. It uses a protected environment called staging/production to adhere to additional protection rules:
    - Enabled: the deploy workflow can only be run from main.
    - To consider: the deploy workflow run has to be approved by a reviewer.

It also prepares for potential future improvement of locking down the staging/production branches and unlocking them only for the force push operation. This is not enabled because this operation would require a more powerful PAT.

## Change checklist


.github/workflows/deploy.yml
```diff
@@ -1,68 +1,81 @@
...
+        description: 'Tag to deploy'
+        required: true
+        type: string
+      environment:
+        description: 'Environment to deploy to'
+        required: true
+        type: choice
+        options:
+          - staging
+          - production
```
lidel: nit: instead of parameter here, we want separate workflows (`deploy-staging.yml` and `deploy-production.yml`). 

this way each has own run history, so its easier to inspect what is deployed where without clicking into each job to see parameter + it is harder to misclick and deploy to wrong place when following release checklist with link to specific workflow.

.github/workflows/deploy.yml
```diff
@@ -1,68 +1,81 @@
...
+            fi
+          fi
+          # If deploying to production, ensure the tag is present in the history of staging
+          if [ "$ENVIRONMENT" == "production" ]; then
+            git merge-base --is-ancestor $TAG origin/staging
+            if [ $? -ne 0 ]; then
+              echo "Tag $TAG is not present in the history of staging"
+              exit 1
+            fi
+          fi
```
lidel: nit: iiuc staging will be same rule as prod, tag should be attached to commit from main branch

SgtPooki:
> To consider: the deploy workflow run has to be approved by a reviewer.

This could be useful but I don't think we need this. Is this even possible?

SgtPooki (approved):
code changes lgtm though i didn't dive deep into the gh api for lock/unlock. is this PR still in draft because we're waiting on something else?

galargh:
It should be ready. I set up a fine-grained token associated with my account and saved it to `staging` and `production` environments.

As for review requirements, you can add that in the environment settings if ever needed.

.github/workflows/deploy.yml
galargh: Good call, I now split them into two. I made them both use a common reusable workflow to avoid duplication.

.github/workflows/deploy.yml
galargh: Got it, thanks. Makes things easier :)

2color:
> To consider: the deploy workflow run has to be approved by a reviewer.

How does this work?

lidel (approved):
lgtm, thanks!

we have small team so anyone with write permission to the repo can deploy, if we see that being a footgun we can figure out more fine-grained permissions, but let's keep things simple  and make a few releases/deployments with this as-is

lidel:
@galargh my attempt to deploy `v1.7.1` to `staging` it failed: https://github.com/ipfs/service-worker-gateway/actions/runs/12183788272/job/33986071546

It seems on-the-fly relaxing of protection did not work?

Perhaps fix is to use a PAT from @web3-bot and then allow force-pushes only to that user here?
> image

galargh:
@lidel That's because both branch protections AND repository rule sets exist for these two. It was https://github.com/ipfs/service-worker-gateway/rules/2510914 which blocked the push. I disabled the ruleset, but left the branch protections in place. Could you try again?

_Edit_: I just tried it myself and it worked - https://github.com/ipfs/service-worker-gateway/actions/runs/12184177003/job/33987425774 🥳

PR #531, opened 2024-12-13
fix: error page merges verified-fetch headers

SgtPooki:
## Change checklist


SgtPooki (commented):
self review

PR #552, opened 2025-01-22
fix: enable verified-fetch server-timing headers

SgtPooki:
## Title

fix: enable verified-fetch server-timing headers

## Description

Passes the `withServerTiming` option to verified-fetch to enable server-timing headers.

## Notes & open questions

Tested locally with `npm run start` and going to http://specs-ipfs-tech.ipns.helia-sw-gateway.localhost/ and checking the server-timing headers in the network tab.

image

## Change checklist


SgtPooki (commented):
self review. note that this PR also updates package-lock.json because I ran `npm run reset && npm i`

PR #566, opened 2025-02-07
fix(ci): timeout and cache playwright deps

lidel:
Applying fix to avoid running for 6h like it did in https://github.com/ipfs/service-worker-gateway/actions/runs/12890491077/job/35940260584 (thx @galargh  for flag)

There is also cache in `.cache/ms-playwright` that is not covered by npm specific cache handling.
Added it so its shared across runs, this way we fetch one version of each browser only once per OS.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,7 +67,16 @@ jobs:
         with:
           node-version: lts/*
       - uses: ipfs/aegir/actions/cache-node-modules@main
+      - name: Cache Playwright dependencies
+        id: playwright-cache
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            playwright-${{ runner.os }}-
       - run: npx playwright install --with-deps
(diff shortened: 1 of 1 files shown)


galargh (approved):

.github/workflows/main.yml
```diff
@@ -67,7 +67,16 @@ jobs:
...
+      - name: Cache Playwright dependencies
+        id: playwright-cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            playwright-${{ runner.os }}-
       - run: npx playwright install --with-deps
+        timeout-minutes: 15 # should take no logner than 2min
```
SgtPooki:
```suggestion
        timeout-minutes: 10 # should take no longer than 2min
```

.github/workflows/main.yml
```diff
@@ -98,7 +107,16 @@ jobs:
...
```
SgtPooki:
```suggestion
        timeout-minutes: 10 # should take no longer than 2min
```

SgtPooki (approved):

PR #601, opened 2025-02-26
feat: add strict CSP policy

2color:
## Description

- This PR explores adding a strict CSP policy to improve the security of the service worker gateway and resilience to attacks. 

## Notes & open questions

- This breaks global config iframe. But maybe this is a desired trade-off? 

## Change checklist

### DIFF
--- a/build.js
+++ b/build.js
@@ -73,7 +73,6 @@ const calculateFileHash = (filePath) => {
  */
 const injectAssets = (metafile) => {
   const htmlFilePath = path.resolve('dist/index.html')
-  const nonce = gitRevision() // Use the git revision as the CSP nonce
 
   // Extract the output file names from the metafile
   const outputs = metafile.outputs
@@ -89,7 +88,7 @@ const injectAssets = (metafile) => {
   console.log(`Script hash (sha256): ${scriptHash}`)
   console.log(`CSS hash (sha256): ${cssHash}`)
 
-  const scriptTag = `<script type="module" nonce="${nonce}" src="${path.basename(scriptFile)}"></script>`
+  const scriptTag = `<script type="module" src="${path.basename(scriptFile)}"></script>`
   const linkTag = `<link rel="stylesheet" href="${path.basename(cssFile)}">`
 
   // Read the index.html file
--- a/build.js
+++ b/build.js
@@ -2,6 +2,7 @@
 import { execSync } from 'node:child_process'
 import fs from 'node:fs'
 import path from 'node:path'
+import crypto from 'node:crypto'
 import esbuild from 'esbuild'
 
 const copyPublicFiles = () => {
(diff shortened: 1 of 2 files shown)


build.js
```diff
@@ -65,12 +79,28 @@ const injectAssets = (metafile) => {
...
   const linkTag = `<link rel="stylesheet" href="${path.basename(cssFile)}">`
 
   // Read the index.html file
   let htmlContent = fs.readFileSync(htmlFilePath, 'utf8')
 
+  // Update the CSP header with the calculated hashes
+  const cspContent = `default-src 'none'; script-src 'self' 'sha256-${scriptHash}'; style-src 'self' 'sha256-${cssHash}'; img-src 'self'; font-src 'self'; manifest-src 'self'; connect-src 'self';`
+  htmlContent = htmlContent.replace(
+    /<meta http-equiv="Content-Security-Policy"[^>]*>/,
+    `<meta http-equiv="Content-Security-Policy" content="${cspContent}">`
```
lidel, line 101: IF we set it here, and not on `ipfs.io` and `dweb.link` as well, we will be sabotaging all efforts to migrate people towards SW-based gateways, because "my website works on old `dweb.link` gateway, and new one `inbrowser.link` does not". 

We had long discussion about CSP in https://github.com/ipfs/in-web-browsers/issues/157 but we never pulled the trigger because PL prioritized `ipfs.io` and `dweb.link`  to "just work", be ok with data exfiltration to support Dapps,  deal with phishing via  [badbits](https://badbits.dwebops.pub/), and then nucleation happened.


:point_right:  I'm happy to lock things down with CSP, it will fix phishing problem, **it will also break most of Dapps talking to remote RPCs or fetching remove subresources etc**.. so if we do it, do it on all domains, not just `inbrowser.tld`.

I don't think we can afford breaking decade old UX on `ipfs.io` gateways, but this is strong belief weakly held, happy to discuss rationale why this is net positive.

lidel (changes requested):
Maybe I misunderstood this PR – what attack are we defending against?

If CSP is applied globally, I don't think it is feasible if we on do it here, and not on `dweb.link` and `ipfs.io`.
Let's discuss below.

build.js
SgtPooki: Doesn't the CSP configuration here only impact the SW gateway build itself? websites loaded with their own index.html should not get this treatment. Or am I missing something?

SgtPooki:
If we break the global config, it will make testing subdomains or users configuring their own local gateway much more painful.. 

the sw-gateway can provide config via JSON with a request now, so maybe we no longer need the iframe?

build.js
lidel: I think it was late when I did first pass :) But.. if this is limited to initial HTML that bootstraps JS, what type of attack this protects us against? Both HTML and JS are fetched from same source, someone can swap HTML instead of JS, and remove CSP, no?

PR #615, opened 2025-03-05
feat: display origin isolation warnings

### LINKED ISSUE #30: Origin isolation warnings, tests and enforcement

SgtPooki:
## Title

feat: show origin isolation warnings

## Description

Fixes #30 

## Notes & open questions

Video demo: 

[2025-03-05 at 10.26.20 - Sapphire Puffin.webm](https://github.com/user-attachments/assets/585f3a48-0ceb-4729-8295-9633c42f1eee)

### Warning in source now looks like this:

image

## Change checklist

### DIFF
--- a/src/pages/subdomain-warning.tsx
+++ b/src/pages/subdomain-warning.tsx
@@ -58,7 +58,6 @@ export default function SubdomainWarningPage (): ReactNode {
   }, [])
 
   useEffect(() => {
-
     if (acceptedRisk) {
       window.location.href = originalUrl ?? '/'
     }
--- a/src/pages/subdomain-warning.tsx
+++ b/src/pages/subdomain-warning.tsx
@@ -1,0 +1,114 @@
+import React, { useCallback, useEffect, useState, type ReactNode } from 'react'
+import Header from '../components/Header.jsx'
+import './default-page-styles.css'
+import './loading.css'
+// import LoadingPage from './loading.jsx'
+
+function IpAddressRecommendations ({ currentHost }: { currentHost: string }): ReactNode {
+  return (
+    <div>
+      <span>Current Host: {currentHost}</span>
+      <p>Ip addresses do not support origin isolation.</p>
+      <p>If you're the website administrator, please ensure your domain has proper DNS configuration</p>
+    </div>
+  )
+}
+
+function DefaultRecommendations ({ currentHost }: { currentHost: string }): ReactNode {
+  return (
+    <div>
+      <span>Current Host: {currentHost}</span>
+      <p>
+        For the best experience, this website should be accessed through a subdomain gateway
+        (e.g., <code>cid.ipfs.{currentHost}</code> instead of <code>{currentHost}/ipfs/cid</code>).
+      </p>
+
+      <p>
+        If you're the website administrator, please ensure your domain has proper DNS configuration
+        for wildcard subdomains (<code>*.ipfs.{currentHost}</code> and <code>*.ipns.{currentHost}</code>).
+      </p>
+    </div>
+  )
+}
+/**
+ * Warning page to display when subdomain setup is not available
+ * This UI is similar to browser security warnings and informs users about missing features
+ */
+export default function SubdomainWarningPage (): ReactNode {
+  const [acceptedRisk, setAcceptedRisk] = useState(sessionStorage.getItem('ipfs-sw-gateway-accepted-path-gateway-risk') != null ?? false)
+
+  const originalUrl = new URL(window.location.href).searchParams.get('helia-sw')
+
+  const handleAcceptRisk = useCallback(async () => {
+    // Store the user's choice in sessionStorage so it persists during the session
+    sessionStorage.setItem('ipfs-sw-gateway-accepted-path-gateway-risk', 'true')
+    // Continue with path gateway by reloading the page
+    // window.location.reload()
+    // post to SW to accept the risk
+    try {
+      await fetch('/#/ipfs-sw-accept-origin-isolation-warning').then(() => {
+        setAcceptedRisk(true)
+      })
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Error accepting risk', error)
+    }
+  }, [])
+
+  useEffect(() => {
+    // eslint-disable-next-line no-console
+    console.log('originalUrl', originalUrl)
+
+    if (acceptedRisk) {
+      window.location.href = originalUrl ?? '/'
+    }
+  }, [originalUrl, acceptedRisk])
+
+  const currentHost = window.location.host
+
+  const isCurrentHostAnIpAddress = /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/.test(currentHost)
+
+  let RecommendationsElement: (props: { currentHost: string }) => ReactNode = DefaultRecommendations
+  if (isCurrentHostAnIpAddress) {
+    RecommendationsElement = IpAddressRecommendations
+  }
+
+  return (
+    <>
+      <Header />
+      <main className='pa4-l bg-yellow-muted mw7 mb5 center pa4 e2e-subdomain-warning mt4'>
+        <div className="flex items-center mb3">
+          <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth="2" strokeLinecap="round" strokeLinejoin="round" className="mr2">
+            <path d="m21.73 18-8-14a2 2 0 0 0-3.48 0l-8 14A2 2 0 0 0 4 21h16a2 2 0 0 0 1.73-3Z"></path>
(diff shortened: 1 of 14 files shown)


src/pages/subdomain-warning.tsx
```diff
@@ -0,0 +1,121 @@
+import React, { useCallback, useEffect, useState, type ReactNode } from 'react'
+import Header from '../components/Header.jsx'
+import './default-page-styles.css'
+import './loading.css'
+import { ServiceWorkerReadyButton } from '../components/sw-ready-button.jsx'
+import { ServiceWorkerProvider } from '../context/service-worker-context.jsx'
+// import LoadingPage from './loading.jsx'
```
SgtPooki:
```suggestion
```

src/pages/subdomain-warning.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const [acceptedRisk, setAcceptedRisk] = useState(sessionStorage.getItem('ipfs-sw-gateway-accepted-path-gateway-risk') != null ?? false)
+  const [isSaving, setIsSaving] = useState(false)
+  const originalUrl = new URL(window.location.href).searchParams.get('helia-sw')
+
+  const handleAcceptRisk = useCallback(async () => {
+    setIsSaving(true)
+    // Store the user's choice in sessionStorage so it persists during the session
+    sessionStorage.setItem('ipfs-sw-gateway-accepted-path-gateway-risk', 'true')
+    // Continue with path gateway by reloading the page
+    // window.location.reload()
```
SgtPooki:
```suggestion

```

src/pages/subdomain-warning.tsx
```diff
@@ -0,0 +1,121 @@
...
+ */
+export default function SubdomainWarningPage (): ReactNode {
+  const [acceptedRisk, setAcceptedRisk] = useState(sessionStorage.getItem('ipfs-sw-gateway-accepted-path-gateway-risk') != null ?? false)
+  const [isSaving, setIsSaving] = useState(false)
+  const originalUrl = new URL(window.location.href).searchParams.get('helia-sw')
+
+  const handleAcceptRisk = useCallback(async () => {
+    setIsSaving(true)
+    // Store the user's choice in sessionStorage so it persists during the session
+    sessionStorage.setItem('ipfs-sw-gateway-accepted-path-gateway-risk', 'true')
```
SgtPooki, line 46: note, we could query the service worker to determine if they've accepted?

src/pages/subdomain-warning.tsx
```diff
@@ -0,0 +1,121 @@
...
+      // eslint-disable-next-line no-console
+      console.error('Error accepting risk', error)
+    } finally {
+      setIsSaving(false)
+    }
+  }, [])
+
+  useEffect(() => {
+    // eslint-disable-next-line no-console
+    console.log('originalUrl', originalUrl)
```
SgtPooki:
```suggestion
```

src/pages/subdomain-warning.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+        <div className="flex justify-center mt4">
+          <ServiceWorkerReadyButton id="accept-warning" label={isSaving ? 'Accepting...' : 'I understand the risks - Continue anyway'} waitingLabel='Waiting for service worker registration...' onClick={() => { void handleAcceptRisk() }} />
+          {/* <ServiceWorkerReadyButton className="f6 link dim br2 ph3 pv2 mb2 dib" id="accept-warning" label={isSaving ? 'Accepting...' : 'I understand the risks - Continue anyway'} waitingLabel='Waiting for service worker registration...' onClick={() => { void handleAcceptRisk() }} /> */}
+          {/* <button
+            className="f6 link dim br2 ph3 pv2 mb2 dib white bg-red pointer"
+            onClick={() => { void handleAcceptRisk() }}
+          >
+            I understand the risks - Continue anyway
+          </button> */}
```
SgtPooki:
```suggestion
```

src/sw.ts
```diff
@@ -42,6 +42,7 @@ interface StoreReponseInCacheOptions {
  */
 interface LocalSwConfig {
   installTimestamp: number
+  originIsolationWarningAccepted: boolean
```
SgtPooki, line 45: this should probably be in private config, like https://github.com/ipfs/service-worker-gateway/blob/957a9506c6c89b069d27ca4f6986e9ee1c6ef4a3/src/lib/config-db.ts#L19-L21

And then we could just query that on the client instead of using sessionStorage

src/sw.ts
```diff
@@ -301,6 +309,10 @@ function isSwConfigGETRequest (event: FetchEvent): boolean {
   return event.request.url.includes('/#/ipfs-sw-config-get')
 }
 
+function isAcceptOriginIsolationWarningRequest (event: FetchEvent): boolean {
+  return event.request.url.includes('/#/ipfs-sw-accept-origin-isolation-warning')
```
SgtPooki, line 313:
```suggestion
  return event.request.url.includes('#/ipfs-sw-accept-origin-isolation-warning')
```

or check for `(new URL(request.url)).hash.startsWith('#/ipfs-sw-accept-origin-isolation-warning')`

src/sw.ts
```diff
@@ -441,6 +454,19 @@ async function fetchHandler ({ path, request, event }: FetchHandlerArg): Promise
         Location: originLocation
       }
     })
+  } else if (isPathGatewayRequest(originalUrl) && !(await getOriginIsolationWarningAccepted())) {
+    const newUrl = new URL(originalUrl.href)
+    newUrl.pathname = '/'
+    newUrl.hash = '/ipfs-sw-origin-isolation-warning'
+    newUrl.searchParams.set('helia-sw', request.url)
+    // sessionStorage.setItem('ipfs-sw-gateway-accepted-path-gateway-risk', 'true')
```
```suggestion

```

no sessionStorage in service worker

SgtPooki (commented):
self review

PR #635, opened 2025-03-18
fix: use default verified-fetch contentTypeParser

SgtPooki:
removes content-type parser in sw-gateway and use the one thats now provided by verified-fetch by default.

related https://github.com/ipfs/helia-verified-fetch/pull/193
related https://github.com/ipfs/helia-verified-fetch/issues/191


SgtPooki (commented):
self review

SgtPooki (commented):
self review

PR #636, opened 2025-03-18
fix: query parameters are preserved

### LINKED ISSUE #628: bug: query param is lost on subdomain after first-hit
1. go to https://inbrowser.link/ipfs/bafybeidh6k2vzukelqtrjsmd4p52cpmltd2ufqrdtdg6yigi73in672fwu/subdir/multiblock.txt?format=car
2. notice redirection to ?helia-sw url
3. sw loads, redirect to subdomain
4. No `format` query param in URL

SgtPooki:
## Title

fix: query parameters are preserved after redirect

## Description

Fixes an issue where query parameters aren't being preserved by IPFS gateways (boxo/rainbow) when redirecting to `?helia-sw=<path>`. Instead of redirecting to `?helia-sw=<path>`, IPFS gateways will return a 200 response pointing to ipfs-sw-first-hit.html, which simply loads `ipfs-sw-first-hit.ts`, which will handle all our redirect logic.

This also required a change to the redirects file to point /* to index.html, which will always load index.html (and all our javascript) for all non /ipfs and /ipns requests.

Fixes #628

## Notes & open questions

cc @lidel lmk if you see anything wrong here, and if we can update cloudflare to work properly with these changes

## Change checklist


package.json
```diff
@@ -16,6 +16,8 @@
     "build:tsc": "tsc",
     "build:esbuild": "node build.js",
     "start": "node build.js --watch --serve",
+    "start:hosted": "run-s build start:with-backend",
+    "start:with-backend": "concurrently --kill-others -c 'auto' -n 'app,gateway,proxy' \"npm run start\" \"cross-env PROXY_PORT=3334 GATEWAY_PORT=8088 POLL_INTERVAL=1000 node test-e2e/ipfs-gateway.js\" \"cross-env BACKEND_PORT=8345 PROXY_PORT=3333 node test-e2e/reverse-proxy.js\"",
```
SgtPooki: I used to do `npm run test:chrome -- --ui` in order to get playwright up and running with all the things we need. Now, we can run `npm run start:hosted` in order to test the sw gateway as if it were in a deployed environment, including ipfs-hosted backend, reverse proxy, etc..

src/ipfs-sw-first-hit.ts
```diff
@@ -0,0 +1,33 @@
...
+  redirectUrl.searchParams.set('helia-sw', path)
+  query.forEach((value, key) => {
+    redirectUrl.searchParams.set(key, value)
+  })
+
+  // eslint-disable-next-line no-console
+  console.log('redirectUrl', redirectUrl.toString())
+
+  // we should remove the history of the redirect to this page
+  // history.replaceState({}, '', redirectUrl.toString())
```
SgtPooki:
```suggestion
```

test-e2e/first-hit.test.ts
```diff
@@ -52,6 +48,18 @@ test.describe('first-hit ipfs-hosted', () => {
...
+
+    test('redirects to ?helia-sw=<path> with extra query params are handled', async ({ page }) => {
+      const response = await page.goto('http://localhost:3334/ipfs/bafkqablimvwgy3y?foo=bar')
+
+      expect(response?.url()).toBe('http://localhost:3334/ipfs/bafkqablimvwgy3y?foo=bar')
+
+      // first loads the root page
+      expect(response?.status()).toBe(200)
+
+      // wait for page to be ?helia-sw=<path>&foo=bar
```
SgtPooki, line 70:
```suggestion
```

test-e2e/first-hit.test.ts
```diff
@@ -52,6 +48,18 @@ test.describe('first-hit ipfs-hosted', () => {
       // and we verify the content was returned
       await page.waitForSelector('text=hello', { timeout: 25000 })
     })
+
+    test('redirects to ?helia-sw=<path> with extra query params are handled', async ({ page }) => {
+      const response = await page.goto('http://localhost:3334/ipfs/bafkqablimvwgy3y?foo=bar')
+
+      expect(response?.url()).toBe('http://localhost:3334/ipfs/bafkqablimvwgy3y?foo=bar')
+
+      // first loads the root page
```
SgtPooki, line 67:
```suggestion
```

test-e2e/ipfs-gateway.js
```diff
@@ -22,93 +22,188 @@ const kuboBin = path()
...
- * @type {import('execa').Options}
+ * @type {import('execa').Options & { env: Record<string, string> }}
  */
 const execaOptions = {
   cwd: __dirname,
   cleanup: true,
   env: {
     IPFS_PATH,
-    GOLOG_LOG_LEVEL: process.env.GOLOG_LOG_LEVEL ?? 'debug,*=debug',
+    GOLOG_LOG_LEVEL: process.env.GOLOG_LOG_LEVEL ?? 'error,*=error',
```
SgtPooki: this was noisy..

test-e2e/ipfs-gateway.js
```diff
@@ -22,93 +22,188 @@ const kuboBin = path()
...
+  }
+
+  if (daemon) {
+    daemon.kill('SIGINT')
+    await new Promise(resolve => daemon.on('exit', resolve))
+    log('Kubo daemon stopped')
+  }
+}
+
+// Function to update IPFS with new dist files
```
SgtPooki:
```suggestion
```

SgtPooki (commented):
self review

public/_redirects
SgtPooki, line 1: @lidel do you see any issues with the approach here? Can we update cloudflare pages to work with the new redirects structure?

Basically we need two/three redirect rules:

1. /ipns/* hidden redirect to /ipfs-sw-first-hit.html
2. /ipfs/* hidden redirect to /ipfs-sw-first-hit.html
3. anything else: index.html

lidel: - Is this specific to path gateways? What about hitting subdomains directly? 
  - `/ipfs/*`  will only engage on `inbrowser.link/ipfs/cid/sub/dir?foo=bar` but not on `cid.ipfs.inbrowser.link/sub/dir?foo=bar`, right?  Or do we handle that already?
- SW registration scoping may be a problem if we explicitly dont make it wider
  - `index.html` `200` for `cid.ipfs.inbrowser.dev/sub/dir/` has to request SW registration for `/`, otherwise it will only handle `/sub/dir/`

Those are only things I see, but if things work locally, the only way to tell for sure is to deploy this to `inbrowser.dev` and see (we have our Nginx, Cloudflare Pages _redirects implementation, and Cloudflare CDN at play).

If we see any issues with `200` responses being cached, we can add below to `<head>` (but let's see if things work without it):

```html
    <meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate">
    <meta http-equiv="Pragma" content="no-cache">
    <meta http-equiv="Expires" content="0">
```

SgtPooki: > Is this specific to path gateways? What about hitting subdomains directly?

We handle this already.

When you hit a subdomain directly, it should return the index.html (because the path requested doesn't exist in the dist dir), which will load the javascript, which will load the service worker, and then handle the path. This logic is all in the `redirect-page.tsx` and we determine whether to load that logic in https://github.com/ipfs/service-worker-gateway/blob/f7526b1e099109b0e7e47da5ad5dee40933e9aef/src/lib/routing-render-checks.ts#L1-L20

> SW registration scoping may be a problem if we explicitly dont make it wider

I actually just confirmed that we have an issue with first-hit for subdomains with a path. We should be registering the sw at the root of the subdomain for sure. I'll open an issue. This is separate from this PR

SgtPooki:
I will update the _redirects file to include explicit identity redirects for all the files in the dist folder (except _redirects)

SgtPooki (commented):
self review

PR #647, opened 2025-03-25
fix: sw registers on root scope

### LINKED ISSUE #646: bug: first-hit on subdomain with path should register service worker at root path
see #645

SgtPooki:
## Title

fix: sw registers on root scope

## Description

This change ensures that the service worker always registers at the root scope of the domain.

In order to do this, any paths in the URL are removed and the helia-sw param is set to the remaining path, then the existing helia-sw logic is applied.

This change also ensures that redirects to subdomains without a path are handled prior to the service worker registering.

Fixes #646

## Change checklist


src/index.tsx
```diff
@@ -21,6 +22,8 @@ if ('serviceWorker' in navigator) {
   ErrorPage = null
 }
 
+void ensureSwScope()
```
SgtPooki: we are doing this in index.tsx to ensure the redirect happens as quickly as possible

src/lib/path-or-subdomain.ts
```diff
@@ -72,6 +72,15 @@ const toSubdomainRequest = (location: Pick<Location, 'protocol' | 'host' | 'path
...
+
+  // create new URL with the subdomain but without the path
+  const newLocation = new URL(`${location.protocol}//${id}.${ns}.${location.host}/`)
+
+  // store the remaining path, existing search params, and hash in the helia-sw param
+  const fullPathToPreserve = `${remainingPath}${location.search}${location.hash}`
+  if (fullPathToPreserve !== '/') {
+    newLocation.searchParams.set('helia-sw', fullPathToPreserve)
+  }
+
```
SgtPooki, line 84: ensure we're redirecting to the subdomain root so we don't have to do so with `ensureSwScope`

src/service-worker-utils.ts
```diff
@@ -5,7 +5,7 @@ import { uiLogger } from './lib/logger.js'
  */
 export async function registerServiceWorker (): Promise<ServiceWorkerRegistration> {
   const log = uiLogger.forComponent('register-service-worker')
-  const swRegistration = await navigator.serviceWorker.register(new URL('ipfs-sw-sw.js', import.meta.url))
+  const swRegistration = await navigator.serviceWorker.register(new URL('ipfs-sw-sw.js', import.meta.url), { scope: '/' })
```
SgtPooki, line 8: ensure scope is set explicitly.

SgtPooki (commented):
self review

PR #649, opened 2025-03-26
test: ipfs-gateway updates CID

SgtPooki:
## Title

fix: ipfs-gateway updates CID

## Description

Currently a draft PR because there are some issues.

Pulled some changes out of https://github.com/ipfs/service-worker-gateway/pull/636

## Notes & open questions

this change attempts to:

1. ensure ipfs-gateway is updated when the dist files are updated
2. ensure IPFS_NS_MAP is updated when the CID changes

## Change checklist

### DIFF
--- a/test-e2e/ipfs-gateway.js
+++ b/test-e2e/ipfs-gateway.js
@@ -22,93 +22,187 @@ const kuboBin = path()
 const gatewayPort = Number(process.env.GATEWAY_PORT ?? 8088)
 
 /**
- * @type {import('execa').Options}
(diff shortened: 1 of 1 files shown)


SgtPooki:
this needs rebased after https://github.com/ipfs/service-worker-gateway/pull/670, but even I sometimes access port 3334 and expect code changes to reflect in queries, and they don't, and then I have to remember to access 3333.. 

At least https://github.com/ipfs/service-worker-gateway/blob/49fe12db8d36a3fb3ba013945277e76efa1d8829/docs/DEVELOPMENT.md now correctly warns against this

PR #652, opened 2025-03-31
fix: first-hit loads config properly

SgtPooki:
## Title

fix: first-hit loads config properly

## Description

Fixes a bug where the config was not loading properly when `ensureSwScope` was called in index.tsx.

## Change checklist


SgtPooki (commented):
self review

PR #654, opened 2025-04-01
chore: dx improvements

SgtPooki:
- **chore: npm start:hosted boots up full test env**
- **test: playwright ipfs-gateway waits for build**


SgtPooki (commented):
self review

PR #673, opened 2025-04-11
feat: single sw-gateway binary

### LINKED ISSUE #629: Distribute Service Worker Gateway as a single binary
## Background

While the SW Gateway handles verification in-browser, you still place trust on the server returning you the correct initial payload of the SW Gateway. This is an attack surface that is hard to overcome in browsers.

## Suggested solution

- Distributing the SW Gateway as a single binary that can be used on localhostc
- The binary can be easily verified with a hash and be published with rigour
- The binary would start a local server serving the files
- origin isolation can be achieved with localhost, e.g. `cid.ipfs.localhost` (as far as I know both Mac and Linux allow this while running in secure context)

## Open questions

- How should we package this into a binary.
-

SgtPooki:
- **feat: build binary and attach to releases**
- **fix: playwright doesnt start webserver if it doesnt need it**
- **test: test local binary**
- **chore: serve-deno.ts is no longer needed**
- **chore: update binary workflow**

## Title

feat: single sw-gateway binary

## Description

In progress

## Change checklist


main.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/http/httputil"
+	"net/url"
+	"sync"
+)
+
+// ----------------------------------------------------------------------------
+// Embedding static assets
+// ----------------------------------------------------------------------------
+
+//go:embed dist/*
```
gammazero, line 18: What is the reason to embed static assets, versus reading the directory at runtime?

SgtPooki: It needs to deploy as a complete binary able to run outside of this repo

test-e2e/config-loading.test.ts
```diff
@@ -43,6 +43,8 @@ test.describe('ipfs-sw configuration', () => {
     })
     const rootConfig = await getConfig({ page })
 
+    // eslint-disable-next-line no-console
+    console.log('trying to go to: ', `${protocol}//bafkqablimvwgy3y.ipfs.${rootDomain}/`)
```
SgtPooki:
```suggestion
```

main.go
```diff
@@ -0,0 +1,95 @@
...
+
+func ipfsLikeHandler(w http.ResponseWriter, r *http.Request) {
+	if fileExists(r.URL.Path) {
+		// Serve the embedded file.
+		distHandler.ServeHTTP(w, r)
+		return
+	}
+	// Otherwise hand off to the SW via redirect.
+	redirectToHelia(w, r)
+}
```
SgtPooki, line 74: note: we could return `ipfs-sw-first-hit.html` here instead, but things are working whether we redirect to `helia-sw` or `ipfs-sw-first-hit.html`

SgtPooki (commented):
self review

SgtPooki:
CI seems stuck so i kicked off https://github.com/ipfs/service-worker-gateway/actions/runs/14626429231

PR #580, opened 2025-02-15
fix: prevent config UI flash on subdomain first load

Winter-Soren:
## Description

When opening a page on a subdomain for the first time, the config UI would flash at the bottom of the page while loading the default index.html in an iframe to read the config. This PR fixes the issue by:

- Adding visibility control for the config iframe
- Only showing iframe after service worker is registered
- Adding loading state coordination between config loading and service worker registration
- Showing loading screen until both config is loaded and service worker is ready

Fixes https://github.com/ipfs/service-worker-gateway/issues/530

## Notes & open questions

The solution improves both the visual experience and loading performance by better coordinating the states between service worker registration and config loading.

Changes are focused on:
- src/pages/redirect-page.tsx

## Change checklist

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation if necessary (this includes comments as well)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #677, opened 2025-04-14
fix: subdomain config loads from root

### LINKED ISSUE #672: bug: more test cleanup
### fix tests using `${protocol}:` 

this is likely rendering a different page than we expect and we need to fix this ASAP because any passing tests are likely false positives.

### fix flaky tests

```
  6 flaky
    [chromium] › test-e2e/byte-range.test.ts:15:3 › byte-ranges › should be able to get a single character 
    [chromium] › test-e2e/byte-range.test.ts:24:3 › byte-ranges › can get 0-0 byte range from car with missing data 
    [chromium] › test-e2e/byte-range.test.ts:33:3 › byte-ranges › can get trailing byte range from car with missing data 
    [chromium] › test-e2e/byte-range.test.ts:44:3 › byte-ranges › video file first set of bytes match kubo gateway 
    [chromium] › test-e2e/hamt-dir.test.ts:13:3 › hamt-dir › can open UnixFS file from HAMT-sharded directory 
    [chromium] › test-e2e/video.test.ts:34:3 › video › starts playing automatically ────────────────
```

SgtPooki:
## Title

fix: subdomain config loads from root

## Description

Fixes an issue with requests to `${protocol}:` that was causing false positives in our tests

fixes the false positives in config-loading.test.ts

Fixes #672
Related #580

## Change checklist


SgtPooki (commented):
self review..

PR #721, opened 2025-05-14
fix: inform users how to deregister the service worker

### LINKED ISSUE #506: Document how to manually deregister SW
> The flag is ipfs-sw-deregister in the URL in any way. see https://github.com/ipfs/service-worker-gateway/blob/fad1eed503bb7ef838274df2b6af1fe9a423a0ac/src/lib/deregister-request.ts#L4

Unsure whats the best place for this.
- README?
- landing page of inbrowser.link itself?
- message in browser console?

Related:
- https://github.com/ipfs/service-worker-gateway/issues/26

SgtPooki:
## Title

fix: inform users how to deregister the service worker

## Description

Fixes #506
Related #26

## Notes & open questions

- Do we want to add a link to the about section so folks can remove it there as well, or is the button at the bottom of the config page sufficient?
- I'm not a big fan of the new deregister button on the config page, but that seems like the most likely place to allow users to deregister the service worker on the root domain and subdomains without having to manually edit the URL.

image
image

## Change checklist


SgtPooki:
updating deregister button to match reset, and move to the left of the save config & reset config buttons to prevent accidental clicks.

image

note that the red only shows on hover

SgtPooki (commented):
self review

README.md
```diff
@@ -121,6 +121,16 @@ aiming to be a drop-in replacement for `ipfs.io`:
 
 Deploying to [production[(https://github.com/ipfs/service-worker-gateway/actions/workflows/deploy-to-production.yml) and [staging](https://github.com/ipfs/service-worker-gateway/actions/workflows/deploy-to-staging.yml) is done by manually running the deployment action and passing the release version to the action.
 
+## Manual Service Worker Deregistration
```
lidel, line 124: nit: could we switch everywhere where we use "deregister" to "unregister" (README, code)?

This way we match Web API (https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/unregister) and avoid using different term.

I dont know if there is a downside or a significant change in meaning, I think "unregister" is fine, just sound less formal? (cc @2color)

src/pages/config.tsx
```diff
@@ -249,8 +249,20 @@ const ConfigPage: FunctionComponent<ConfigPageProps> = () => {
           />
         </InputSection>
         <div className="w-100 inline-flex flex-row justify-between">
-          <button className="e2e-config-page-button button-reset mr5 pv3 tc bg-animate hover-bg-gold pointer w-30 bn" id="reset-config" onClick={() => { void doResetConfig() }}>Reset Config</button>
-          <ServiceWorkerReadyButton className="e2e-config-page-button white w-100 pa3" id="save-config" label={isSaving ? 'Saving...' : 'Save Config'} waitingLabel='Waiting for service worker registration...' onClick={() => { void saveConfig() }} />
+          <ServiceWorkerReadyButton
+            className="e2e-config-page-button pv3 tc bg-animate hover-bg-red-muted pointer w-30 bn"
+            id="deregister-sw"
+            label="Deregister SW"
+            waitingLabel='Waiting for SW...'
```
lidel: nit: "Reset Worker"  is shorter and will look better next to "Reset Config"

```suggestion
            label="Reset Worker"
            waitingLabel='Waiting for SW...'
```

src/components/About.tsx
```diff
@@ -8,7 +8,7 @@ export default function About (): ReactElement {
...
           <p><strong>Is this production ready?</strong> This project is under heavy development and is not yet fully compliant with <a href="https://specs.ipfs.tech/http-gateways/" target="_blank">IPFS Gateway Specfications</a>. Track our efforts  <a href="https://github.com/ipfs/service-worker-gateway/milestones" target="_blank">here</a>.</p>
-          <p><strong>Found a bug?</strong> We welcome you to report it by <a href="https://github.com/ipfs/service-worker-gateway/issues/new" target="_blank">opening an issue</a> with details like an address and a screenshot.</p>
+          <p><strong>Found a bug?</strong> We welcome you to report it by <a href="https://github.com/ipfs/service-worker-gateway/issues/new" target="_blank">opening an issue</a> with details like an address and a screenshot. If you need to manually deregister the service worker, you can do so by appending <code>?ipfs-sw-deregister=true</code> to the URL.</p>
```
lidel: nit: good place to link to exact Web API function that will get called:

```suggestion
          <p><strong>Found a bug?</strong> We welcome you to report it by <a href="https://github.com/ipfs/service-worker-gateway/issues/new" target="_blank">opening an issue</a> with details like an address and a screenshot. If you need to manually <a href="https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/unregister" target="_blank">unregister</a> the service worker, you can do so by appending <code>?ipfs-sw-unregister=true</code> to the URL.</p>
```

README.md
SgtPooki: yea we can switch to unregister so if people search it up it's easy to find. I don't necessarily agree with using the same terms as APIs.. but both terms mean the same and it's not necessarily jargon

SgtPooki:
@lidel One other question here: Since the unregister is currently redirecting to the config page, how do we want to handle "unregister" going forward? Should we redirect at all? `window.reload`? do we want to handle it differently for subdomains?



EDIT: I removed the redirect.. users will need to manually refresh after unregister.

PR #738, opened 2025-05-28
fix: update deps and ensure 504 test passes

SgtPooki:
Updates helia and libp2p deps, ensures 504 timeout test is passing.


SgtPooki (commented):
self review

PR #751, opened 2025-06-11
fix: provide content-renderer UI

SgtPooki:
## Title

fix: provide content-renderer UI

## Description

Adds a content-viewer for Safari browsers. This safari targeted fix is no longer required but a content-renderer may help with getting around the chromium bug documented at #574.

Related https://github.com/ipfs/service-worker-gateway/issues/574
Related https://github.com/ipfs/service-worker-gateway/pull/753

## Change checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation if necessary (this includes comments as well)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #753, opened 2025-06-12
fix: safari renders images correctly

SgtPooki:
## Title

fix: safari renders images correctly

## Description

Safari was doing something weird with the response coming from verified-fetch, simply by constructing a new response object with the same body, status, statusText, and headers was enough to fix the issue.

Related: https://github.com/ipfs/service-worker-gateway/pull/751
Related: https://github.com/SgtPooki/safari-sw-image-bug-repro

## Notes & open questions

When comparing the responses with the following function:

```
function summarizeResponseDifferences(r1: Response, r2: Response): string {
  const diff: string[] = []

  const fieldsToCheck = [
    'status',
    'statusText',
    'ok',
    'redirected',
    'type',
    'url',
    'bodyUsed'
  ]

  for (const key of fieldsToCheck) {
    const v1 = r1[key]
    const v2 = r2[key]
    if (v1 !== v2) {
      diff.push(`${key}: "${v1 ?? 'null'}" (old) -> "${v2 ?? 'null'}" (new)`)
    }
  }

  // Headers diff
  const headers1: Record<string, string> = {}
  const headers2: Record<string, string> = {}
  r1.headers.forEach((v, k) => { headers1[k] = v })
  r2.headers.forEach((v, k) => { headers2[k] = v })

  for (const key of new Set([...Object.keys(headers1), ...Object.keys(headers2)])) {
    if (headers1[key] !== headers2[key]) {
      diff.push(`header:${key}: "${headers1[key] ?? 'null'}" (old) -> "${headers2[key] ?? 'null'}" (new)`)
    }
  }

  return diff.join('; ')
}
```

The only observed differences were:

| Property | New Value                                                                                         | Old Value |
|----------|---------------------------------------------------------------------------------------------------|-----------|
| `type`   | `default`                                                                                         | `basic`   |
| `url`    | *(empty)*                                                                                         | `http://127.0.0.1:3000/ipfs/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi` |

## Change checklist


SgtPooki (commented):
self review

Issue #687, opened 2025-04-22
fix: update loading text and screen on redirect-page


SgtPooki:
https://github.com/ipfs/service-worker-gateway/blob/43677c48a728f4b9f42e13f57ca4f438427e3dba/src/pages/redirect-page.tsx#L85-L94

we shouldn't mention "auto reload" or anything like that.. we should also probably just use `<LoadingIndicator />`

dhvll:
I would like to work on this issue. Could you assign it to me?

SgtPooki:
@dhvll sure, thank you!

SgtPooki:
@dhvll any update?

dhvll:
Yes will send by EOD

lidel:
Related: https://github.com/ipfs/service-worker-gateway/issues/530#issuecomment-2974004117

lidel:
redirect-page no longer exist – adjustment to initial UX should be filled as new issue

Issue #530, opened 2024-12-13
Config flashes before "Loading" screen on subdomains


lidel:
When opening page on a subdomain for the first time, config flashes at the bottom of the page:

> [screen-recorder-fri-dec-13-2024-17-29-57.webm](https://github.com/user-attachments/assets/899f9c22-5489-4ba6-ae0a-0687f7d89c9a)

This likely occurs when default index.html loads in iframe to read config. 
We should hide entire DOM if URL is landing page (no path, no subdomain).

Winter-Soren:
Can you assign me this issue, I would love to resolve this.

SgtPooki:
@Winter-Soren please feel free to work on this and open a PR :)

Winter-Soren:
@SgtPooki I've made changes to fix the config UI flash issue on subdomain first load (#530). Would appreciate your review and guidance to ensure I'm approaching this in the right direction. The solution focuses on managing iframe visibility and loading states in redirect-page.tsx.

lidel:
This seems to be happening again on inbrowser.link and .dev.

Let's take an opportunity to  improve the UI.

Good idea from @autonome:

> would be super cool if that interstitial had all website ui removed, just be a loading animation in the middle, with a very short msg communicating the value of the wait

related ideas:
- reuse fullscreen animation from the header of  [helia.io](https://helia.io/)?  
- reuse heartbeat css from ipfs-companion?
- have text that says "initializing Service Worker... TIP: visit `inbrowser.link` to adjust settings" ?

SgtPooki:
@lidel We *have* to show the iframe or it wont load config from global domain. do we still want that functionality?

config iframe started showing again in https://github.com/ipfs/service-worker-gateway/pull/677

SgtPooki:
One thing we can do with config loading from root that I forgot about until just now was the get config JSON call to the SW.. instead of loading the UI (dealing with defaults, async loading issues, and UI delays, etc.) the iframe config should simply query the SW-config `await fetch('?ipfs-sw-config-get=true')` and post that to the subdomain page (the parent of the iframe)..

Can we return an HTML page that simply handles "postMessage" directly from the service worker...? e.g.

```
<iframe src=`${parentDomain}?ipfs-sw-config-get-for-subdomain=true` />
```

2color:
> We have to show the iframe or it wont load config from global domain

[...]

PR #759, opened 2025-06-18
fix: global config is passed to subdomains

### LINKED ISSUE #757: Research config sharing that works with Storage Partitioning
## Problem

More and more browsers seems to apply Storage Partitioning, which breaks the way we are sharing config set on `inbrowser.link` with all subdomains.

Filling this issue as a pointer for work related to finding workaround / alternative way of sharing global config with each subdomain init.

cc https://github.com/ipshipyard/roadmaps/issues/5

## Ideas to explore

- switch to URL-based config passing (`cid.ipfs.inbrowser.link`→`inbrowser.link?path`→`cid.ipfs.inbrowser.link?config`→`cid.ipfs.inbrowser.link`)

### Reference

- Firefox: https://developer.mozilla.org/en-US/docs/Web/Privacy/Guides/State_Partitioning#dynamic_partitioning
- Chromium: https://privacysandbox.google.com/cookies/storage-partitioning
- https://storage-partitioning-demo-site-a.glitch.me/

SgtPooki:
## Title

fix: global config is passed to subdomains

## Description

This PR changes the way we handle subdomain config loading (Fixing it, because it's currently broken, see #757), resulting in a much faster pull from the root domain the service worker gateway is on by doing a little redirect dance between the root config and the subdomain.

Basic flow:

1. land on subdomain
2. redirect to root with query param asking to get config
3. root pulls config, compresses using lz-string into a query param (default config is a 396 char string, so should be plenty of room for other config), and then redirects back to the subdomain with that query parameter, and a merged helia-sw parameter (helia-sw may be provided by our main.go, cloudflare, ipfs host via redirects file, etc. that currently point to `ipfs-sw-first-hit.html`)
4. index.html hits on subdomain, config is recognized in url, pulled, decompressed and set in IndexedDB, service worker is registered, and we redirect to the proper origin isolated URL with any remaining path.

This PR also changes how the initial JS is loaded so that we are not bundling so much code into the initial page. The entire flow (1-4) above requires very minimal javascript and our unchanged index.html page (that used to load up the entire react app), which is all rendering no UI, so there are no flashes of UI while the redirects are happening (see #530).

Also, because the config is loaded in 4 above before the service worker is ever registered, loads are much quicker as well.

Related https://github.com/ipfs/service-worker-gateway/issues/530
Related https://github.com/ipfs/service-worker-gateway/issues/687
Related https://github.com/ipfs/service-worker-gateway/pull/760
Fixes https://github.com/ipfs/service-worker-gateway/issues/757

## Notes & open questions

Things to still do:

- [ ] Add more tests for first-hit-helpers
- [ ] Remove any old iframe code (mostly in tests now)
- [x] Investigate passing the config to the subdomain using a hashfragment instead of a query parameter - see https://github.com/ipfs/service-worker-gateway/pull/760

## Change checklist


src/index.tsx
```diff
@@ -1,46 +1,47 @@
...
-  { shouldRender: renderChecks.shouldRenderSubdomainWarningPage, component: LazySubdomainWarningPage },
-  { shouldRender: async () => renderChecks.shouldRenderRedirectsInterstitial(), component: LazyInterstitial },
-  {
-    shouldRender: async () => renderChecks.shouldRenderRedirectPage(),
-    component: LazyRedirectPage
+async function main (): Promise<void> {
+  const url = new URL(window.location.href)
+  const state = await getStateFromUrl(url)
+
+  if (state.subdomainHasConfig && state.isIsolatedOrigin) {
```
SgtPooki: comment how this will show only after redirect dance, and only once. When the SW is registered, this should never run because queries will be captures.

src/lib/config-db.ts
```diff
@@ -215,3 +220,34 @@ export async function areSubdomainsSupported (logger?: ComponentLogger): Promise
...
+  const timestamp = Date.now()
+  const configJson = JSON.stringify({ config, timestamp })
+
+  return LZString.compressToEncodedURIComponent(configJson)
+}
+
+export async function decompressConfig (compressedConfig: string): Promise<ConfigDb> {
+  const { config, timestamp } = JSON.parse(LZString.decompressFromEncodedURIComponent(compressedConfig))
+  // if the config is more than 10 seconds old, throw an error
+  if (timestamp < Date.now() - 1000) {
```
SgtPooki: 1 second delay makes it very challenging to inject configs into users' subdomains, but I'm sure we could be a little more robust here.

cc @lidel

test/path-or-subdomain.spec.ts
```diff
@@ -1,49 +1,169 @@
...
+
+      const out = toSubdomainRequest(loc)
+      const url = new URL(out)
+      expect(url.origin).to.equal(`http://foo-bar.ipns.${hostname}`)
+      expect(url.pathname).to.equal('/')
+      expect(url.searchParams.get('helia-sw')).to.equal('/baz')
+    })
+
+    it('ignores invalid namespaces', () => {
+      // TODO: This test was added without modifying the code, and i'm not sure we want this functionality.
```
SgtPooki, line 162: callout for attention here. there are checks and guards elsewhere that ensure users aren't being redirected to this actual domain if its not ipfs/ipns.. but loading https://inbrowser.dev/potato/QmWhatever/foo results in the landing page, and we should probably have some error screen for this

src/sw.ts
```diff
@@ -203,7 +204,10 @@ async function requestRouting (event: FetchEvent, url: URL): Promise<boolean> {
       status: 200
     }))
     return false
-  } else if (isConfigPageRequest(url)) {
+  } else if (isSubdomainConfigRequest(event)) {
+    log.trace('subdomain config request, ignoring and letting index.html handle it', event.request.url)
```
SgtPooki, line 208: note that this should never happen.. service worker isn't registered until after all the redirect dance is handled.

We may want to investigate migrating all the redirect logic added to index.tsx to the service worker.

src/lib/get-subdomain-parts.ts
```diff
@@ -7,10 +7,11 @@ export interface UrlParts {
 }
 
 export function getSubdomainParts (urlString: string): UrlParts {
-  const labels = new URL(urlString).hostname.split('.')
+  const url = new URL(urlString)
+  const labels = url.host.split('.')
   let id: string | null = null
   let protocol: string | null = null
-  let parentDomain: string = urlString
+  let parentDomain: string = url.host
```
SgtPooki, line 14: fixes an issue where parentDomain is not set correctly if its not a legitimate subdomain. there are tests added for this.

SgtPooki (commented):
self review

test/path-or-subdomain.spec.ts
SgtPooki: opened https://github.com/ipfs/service-worker-gateway/issues/772 to track

src/lib/config-db.ts
SgtPooki: FYI timestamp check here was wrong but it's been fixed in [`82075e3` (#759)](https://github.com/ipfs/service-worker-gateway/pull/759/commits/82075e3dadcc70a757d7a48efe652e3d3adc992c)

PR #760, opened 2025-06-19
feat: migrate to hash fragments

SgtPooki:
This codechange is on top of https://github.com/ipfs/service-worker-gateway/pull/759 and shows that we can use hashfragments instead of query parameters, meaning servers & our service worker are not queried. This saves us some roundtrips and network hops, and prevents the config from being passed to backend servers.

Related https://github.com/ipfs/service-worker-gateway/pull/759


SgtPooki (commented):
self review.. everything passing locally

PR #773, opened 2025-06-23
fix: do not import css or react JS unless needed

SgtPooki:
## Title

fix: do not import css unless needed

## Description

Dynamically import CSS and app.tsx JS file only when we need to render the UI instead of on every page load.

For content-addressed data requests (the whole point of the service worker gateway), we don't need to load the CSS or react JS.

Related https://github.com/ipfs/service-worker-gateway/pull/759

## Notes & open questions

The current CSS file is ~90kb, and an extra network request is made for every page load (2 non-cached requests when hitting a subdomain directly). Removing this should help with the initial load time.

I’ve got CSS and js for react/ui code importing dynamically now and the initial bundle size has been reduced (after redirect-config bounce changes in https://github.com/ipfs/service-worker-gateway/pull/759) from 255kB to 39kB for the initial javascript.. and removes 90~kb of CSS on index.html page loads.

Basically this results in a significant first-load reduction from ~32s _using 3g throttling_ to 19s in chrome. (using “finish” time in chrome dev tools.. i.e. from the time all the network requests started (hey, gimme the content of this URL) to redirect+style+content loading + rendering being done)

without the throttling, the reduction in loadtime is improved but almost insignificant. (load time 857ms, DOMContentLoaded 476ms) — not much difference from https://github.com/ipfs/service-worker-gateway/pull/759

## Change checklist


tsconfig.json
```diff
@@ -16,7 +16,7 @@
   },
   "include": [
     "src",
-    "types/svg.d.ts",
+    "types",
```
SgtPooki:
```suggestion
    "types/svg.d.ts",
```

test/node.ts
```diff
@@ -46,7 +46,7 @@ describe(`verify-dist at ${distRoot}`, () => {
     {
       filePath: 'index.html',
       content: [
-        /ipfs-sw-index-.*\.css/,
+        // /ipfs-sw-index-.*\.css/, // no longer needed, css is injected dynamically
```
SgtPooki:
```suggestion
```

SgtPooki (commented):
self review. tested locally and works fine.

no FOUC, but it's also technically a race condition ¯\_(ツ)_/¯

src/index.tsx
```diff
@@ -1,12 +1,29 @@
...
+
+    const script = document.createElement('script')
+    script.type = 'module'
+    script.src = `/${APP_FILENAME}`
+    //     script.textContent = `
+    // import renderApp from '/${APP_CHUNK_FILENAME}';
+    // renderApp();
+    // console.log('import.meta.url', import.meta.url);
+    // export {}
+    // `
```
SgtPooki:
```suggestion
```

SgtPooki (commented):
self review after adding dynamic JS loading and splitting react/ui JS out from initial page hits

PR #776, opened 2025-06-23
fix: cloudflare redirect not being handled properly

SgtPooki:
- **fix: cloudflare redirect works**

Related https://github.com/ipfs/service-worker-gateway/pull/775


SgtPooki (commented):
self review

PR #785, opened 2025-06-24
fix: remove unused redirect-page

SgtPooki:


SgtPooki (commented):
self review. only removals. all tests pass locally

PR #791, opened 2025-06-26
fix: allow injection of trustless config

### LINKED ISSUE #790: bug: invalid config results in user stuck on loading screen
Image

we should fail silently and use the default config

SgtPooki:
- **fix: allow injection of trustless config**
- **test: ensure injected config works as expected**

## Title

fix: allow injection of trustless config

## Description

Allows injection of config that is considered untrusted, but only overrides settings that are allowed to be overridden.

We do not allow injection of URLs if the config is considered untrusted.

See discussion at https://github.com/ipfs/helia/discussions/824 for more details

## Notes & open questions

This PR also removes the dependency on lz-string, reducing initial javascript loaded by about 10kb, but increases size of the `ipfs-sw-cfg` content by about 200 bytes. This should result in a net reduction in bandwidth usage.

* bandwidth used *was* +10kb for every new subdomain + 10kb for uncached root domain js.
* bandwidth used *is* now _reduced_ by 10kb-200bytes for every new subdomain after redirect dance, -10kb for any uncached root domain js

## Change checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,6 @@
         "ipfs-css": "^1.4.0",
         "ipfsd-ctl": "^15.0.2",
         "libp2p": "^2.8.12",
-        "lz-string": "^1.5.0",
         "multiformats": "^13.3.2",
         "react": "^19.0.0",
         "react-dom": "^19.0.0",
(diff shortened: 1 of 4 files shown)


SgtPooki (commented):
self review

2color:
@SgtPooki Any chance you can document how to inject config, e.g. disabling fallback on recursive gateways? 

This came up in a discussion with @dennis-tra

PR #832, opened 2025-08-15
fix: SVG not visible issue

### LINKED ISSUE #823: SVG image not loading on website
I've discovered a small issue where an SVG image is not showing up when loading an ENS based website through the service worker gateway.

On the url: https://simplepage-eth.ipns.inbrowser.link/ you can observe that the "Logo" top left isn't loaded. Also the favicon isn't loaded as well. Both of these are set to [`logo.svg`](https://simplepage-eth.ipns.inbrowser.link/_assets/images/logo.svg) (which does load), but for some reason it's not showing up on the page.

For reference this works fine on https://simplepage.eth.link, and if you load it through the kubo gateway directly.

SAHU-01:
## Title
fix(sw): serve .svg as image/svg+xml

## Description

This PR fixes https://github.com/ipfs/service-worker-gateway/issues/823

**_Problem_**

- **_SVG not rendering on *.inbrowser.link because the SW returned Content-Type_**: application/xml for .svg. Browsers won’t paint SVGs in  / favicon when the MIME is XML, so the logo and favicon were invisible.

- _**Build failure on strict TS (TS2345):**_
ReactDOMClient.createRoot(container) where container was HTMLElement | null. On stricter type defs for @types/react-dom, createRoot requires a non-null Container. (See screenshot in the issue.)
image

**_Changes Implemented_**

- **_Content-Type override for SVG in sw.ts:_**
Force image/svg+xml for .svg/.svgz (independent of Accept), add X-Content-Type-Options: nosniff, and ensure the corrected headers are used in the final Response.

- **_Cache bust:_**  `const CACHE_VERSION = 2.`
Why needed: many clients would have the old cached responses for /…/logo.svg with the wrong Content-Type. Bumping the version rotates cache names; the activate handler deletes old caches so users pick up the corrected headers immediately without manual hard-refresh.

- **_Root element null-guard in src/app.tsx:_** fixes the TS2345 error shown in the attached screenshot by guaranteeing Container is non-null on all environments.

Now the SVG's are visible
image

## Change checklist

- [X] I have performed a self-review of my own code
- [X] I have made corresponding changes to the documentation if necessary (this includes comments as well)
- [X] I have added tests that prove my fix is effective or that my feature works


SgtPooki:
fyi @SAHU-01 the svg issue should be fixed in https://github.com/ipfs/helia-verified-fetch. I just pushed up https://github.com/ipfs/helia-verified-fetch/pull/269

SgtPooki:
also, the typescript errors aren't happening for me because we were missing types.. fixing that up (with other type errors) in https://github.com/ipfs/service-worker-gateway/pull/852

Thanks so much for getting on this fix and pushing up a PR, but I think we can close this one now

SgtPooki:
this was fixed and is live on inbrowser.dev. thanks for your work looking into this fix and getting a PR out

Issue #840, opened 2025-09-05
feat: focus on safe, compliant, local-first, non-expiring service worker


SgtPooki:
e.g. remove serviceWorkerRegistrationTTL/timebomb/expiry... a continuation of the original thought behind https://github.com/ipfs/service-worker-gateway/issues/724, without as much verbal communication of the issue, and more documentation of my reasoning.

We chatted about this at [yesterday's Helia WG](https://github.com/ipfs/helia/discussions/847) and so I wanted to come back to this with a more thorough argument (that I've voiced before, but never written down).

So.. built some logic tables to map out how service worker updates are actually provided to the user. The tables below compare service worker update scenarios with and without the registration TTL/timebomb mechanism.

# Service Worker Update Logic Tables

This was generated with the help of an LLM, and then cleaned up and verified. The summary section at the bottom is 99% human-generated.

## Variables

- **userOnlineStatus**: Online/Offline
- **isSwUpdateAvailable**: Yes/No (new version exists)
- **userNavigationBehavior**: Navigates/Stays on same page
- **timeSinceLastUpdateCheck**: < 24h / ≥ 24h (since last update check that browsers do) see [FF](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#updating_your_service_worker:~:text=If%20your%20service%20worker%20has%20previously%20been%20installed%2C%20but%20then%20a%20new%20version%20of%20the%20worker%20is%20available%20on%20refresh%20or%20page%20load%2C%20the%20new%20version%20is%20installed%20in%20the%20background) and [Chromium](https://developer.chrome.com/docs/workbox/service-worker-lifecycle#when_updates_happen)
- **ttlExpired**: Yes/No (TTL has expired, only relevant when TTL is used)
- **doesUserGetUpdatedSw**: Yes/No (final outcome)

## Table 1: WITHOUT TTL (Pure Browser Updates)

| Scenario | Online | Update Available | Navigation | Time Since Check | Gets Updated SW | Notes |
|----------|--------|------------------|------------|------------------|-----------------|-------|
| 1 | ✅ | ✅ | ✅ | < 24h | ✅ | Normal update on navigation |
| 2 | ✅ | ✅ | ✅ | ≥ 24h | ✅ | Update check triggered after 24h |
| 3 | ✅ | ✅ | ❌ | < 24h | ❌ | No navigation = no update check |
[...]

PR #841, opened 2025-09-05
fix: refresh and ttl edgecases

### LINKED ISSUE #822: index.html shows on subdomain on ctrl+shift+r
Upon loading a subdomain url, e.g. https://bafybeibgua76y24qarr6gpagfr67qzsyutpo2otybsqqgnsk4xe4ur5wmy.ipfs.inbrowser.link/ I got the root config page

I'm not 100% certain why, but upon discussion with @SgtPooki, it seems that this might be related to the "time bomb"

SgtPooki:
- **fix: hard refresh uses service-worker**
- **test: remove unused FrameLocator types**
- **fix: registrationTTL unregisters sw after returning response**

## Title

fix: refresh and ttl edgecases

## Description

Fixes #822

## Notes & open questions

I still think we need to work towards removing the registrationTTL, and work on implementing badbits. I threw some more thoughts down at https://github.com/ipfs/service-worker-gateway/issues/840.

## Change checklist

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -49,9 +49,11 @@ async function main (): Promise<void> {
       await ensureSwScope()
       await registerServiceWorker()
     } catch (err) {
+      // eslint-disable-next-line no-console
       console.error('helia:sw-gateway:index: error ensuring sw scope and registration', err)
     }
     window.location.replace(translatedUrl.href) // replace with translated URL to remove helia-sw param
+    await renderUi()
     return
   }
 
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -23,7 +23,7 @@ async function renderUi (): Promise<void> {
 
 async function main (): Promise<void> {
   if (!('serviceWorker' in navigator)) {
-    // no service worker, render the UI
+    // no service worker support, render the UI
     await renderUi()
     return
   }
@@ -36,16 +36,35 @@ async function main (): Promise<void> {
     await renderUi()
     return
   } else if (state.hasConfig) {
+    // user is requesting content addressed data and has the config already
+    // we need to ensure a SW is registered and let it handle the request
+    const translatedUrl = translateIpfsRedirectUrl(url)
     if (state.isHardRefresh) {
-      // this is a hard refresh, we need to soft-reload to ensure the service worker captures the request.
-      window.location.reload()
+      // this is a hard refresh, we need to reload to ensure the service worker captures the request.
+      window.location.replace(translatedUrl.href) // replace with translated URL to remove helia-sw param
       return
     }
-    // user is requesting content addressed data and has the config already, render the UI
-    await renderUi()
+    // else, there is some other reason why sw didn't capture the request, ensure sw is registered and reload
+    try {
+      await ensureSwScope()
+      await registerServiceWorker()
+    } catch (err) {
+      console.error('helia:sw-gateway:index: error ensuring sw scope and registration', err)
+    }
+    window.location.replace(translatedUrl.href) // replace with translated URL to remove helia-sw param
     return
   }
-  // the user is requesting content addressed data and does not have the config, continue with the config loading flow
+
+  /**
+   * **********************************************
+   * From here on, we are handling the case where:
+   * - the user is requesting content addressed data
+   * - the user does not have the config
+   *
+   * We need to load the config, which may involve a redirect to the root domain
+   * before the service worker can be registered
+   * ***********************************************
+   */
   loadingIndicatorElement?.classList.remove('hidden')
 
   if (state.hasConfig && state.isIsolatedOrigin) {
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -31,11 +31,16 @@ async function main (): Promise<void> {
   const url = new URL(window.location.href)
   const state = await getStateFromUrl(url)
 
-  if (!state.requestForContentAddressedData) {
+  if (!state.requestForContentAddressedData || state.isConfigRequest) {
     // not a request for content addressed data, render the UI
     await renderUi()
     return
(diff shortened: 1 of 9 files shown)


src/index.tsx
```diff
@@ -31,11 +31,16 @@ async function main (): Promise<void> {
...
     await renderUi()
     return
   } else if (state.hasConfig) {
+    if (state.isHardRefresh) {
+      // this is a hard refresh, we need to soft-reload to ensure the service worker captures the request.
+      window.location.reload()
+      return
+    }
     // user is requesting content addressed data and has the config already, render the UI
     await renderUi()
```
SgtPooki: we should likely never need to do this.. I think we can remove it. if they have the config and it's a request for contentaddresseddata (pathRequest|subdomainRequest), then we should be calling `ensureSw` and `loadServiceWorker` and reloading..? if this main page JS is running, and we are requesting content, and the service worker is registered, it should have captured the request.

we need to address this

test-e2e/fixtures/locators.ts
```diff
@@ -1,10 +1,7 @@
-import type { FrameLocator, Locator, Page } from '@playwright/test'
```
SgtPooki, line 1: some cleanup

src/sw.ts
```diff
@@ -201,11 +216,7 @@ self.addEventListener('fetch', (event) => {
  ******************************************************
  */
 async function requestRouting (event: FetchEvent, url: URL): Promise<boolean> {
-  if (!isServiceWorkerRegistrationTTLValid()) {
-    log.trace('Service worker registration TTL expired, unregistering service worker')
-    event.waitUntil(self.registration.unregister())
-    return false
-  } else if (isUnregisterRequest(event.request.url)) {
```
SgtPooki, line 208: moved to after the request to prevent weird network failure issues because sw unregisters in the middle of a response

src/lib/first-hit-helpers.ts
```diff
@@ -154,10 +164,28 @@ export async function getStateFromUrl (url: URL): Promise<NavigationState> {
...
   if (isIsolatedOrigin) {
     // check if indexedDb has config
     hasConfig = await isConfigSet(uiLogger)
+    if (hasConfig) {
+      // Check service worker state
+      const registration = await navigator.serviceWorker.getRegistration()
+      const hasActiveWorker = registration?.active != null
+      const hasControllingWorker = navigator.serviceWorker.controller != null
+
+      if (hasActiveWorker && !hasControllingWorker) {
```
SgtPooki, line 179: in chrome & firefox, hasActiveWorker will be true and hasControllingWorker will be false after a hard refresh.

in playwright-firefox, this is not true. it may be an out of date browser runtime?

SgtPooki (commented):
self review

src/index.tsx
```diff
@@ -31,11 +31,16 @@ async function main (): Promise<void> {
...
-  if (!state.requestForContentAddressedData) {
+  if (!state.requestForContentAddressedData || state.isConfigRequest) {
     // not a request for content addressed data, render the UI
     await renderUi()
     return
   } else if (state.hasConfig) {
+    if (state.isHardRefresh) {
+      // this is a hard refresh, we need to soft-reload to ensure the service worker captures the request.
+      window.location.reload()
+      return
```
lidel: 💭  not sure how real risk is, but to avoid infinite reload loop we may want to do something like

```js
// Prevent reload loops
const reloadKey = 'sw-hard-refresh-reload'
if (sessionStorage.getItem(reloadKey)) {
  sessionStorage.removeItem(reloadKey)
  // Already reloaded once, don't loop
  return
}
sessionStorage.setItem(reloadKey, '1')
  window.location.reload()
  return
```

src/sw.ts
```diff
@@ -188,7 +188,22 @@ self.addEventListener('fetch', (event) => {
...
+          setTimeout(() => {
+            // we unregister after a timeout to avoid interfering with the response.
+            // This solves some edge cases where the response is not complete yet, and we unregister the service worker
+            // while a response is in-flight.
+            // Ideally, we would return a temporary redirect response, and deregister the service worker between the
+            // time it took to return the response and the time it took to request the redirect URL, but it seems like
+            // that will need some investigation or some changes on the client side to support.
+            // Even more ideally, we would not have a registrationTTL to begin with, as the browser handles sw updates automatically.
+            self.registration.unregister()
+          }, 50)
```
lidel: this feels racy, could we wait for response to arrive?  or .. does it have to be as low as 50?

maybe:

```js
event.respondWith(
  getResponseFromCacheOrFetch(event).then(async (response) => {
    if (!isServiceWorkerRegistrationTTLValid()) {
      const clonedResponse = response.clone()
      event.waitUntil(
        clonedResponse.blob().then(() => {
          log.trace('Service worker registration TTL expired, unregistering after response consumed')
          return self.registration.unregister()
        })
      )
      return response;
    }
    return response;
  })
)
```

lidel (commented):
thanks, lets do whatever we need to fix #822 asap.

if you feel this is the way, sgtm, but i'm also ok with removing timebomb entirely if that makes things easier / less brittle

src/sw.ts
SgtPooki: it's racy for sure.. I will try this.. and if I can't get it working, then based on your response in #840 I think we should remove the TTL.. I can leave it commented out for now in case we need to re-add it back

src/index.tsx
SgtPooki: I don't think there is a risk because one flag is whether or not the service worker is registered. isHardRefresh is only true if there is an active worker, and the controller for that worker is null.. a reload of the page means the service worker (that we already confirmed exists) will capture it

PR #885, opened 2025-10-18
fix: broken cache

### LINKED ISSUE #861: bug: service-worker cache seems to be broken
No matter how many times I refresh a page, I cannot get `cached response HIT` to show in the logs. We should probably be using a persistent blockstore instead and this would solve most of our problems.

I believe cache.put(cacheKey, resp) and `cache.match(cacheKey)` will not work because of the way the browser cache works... but I also think we have a bug in some mutable checks..

also, fetchAndUpdateCache is always passing mutable=true with ` await storeReponseInCache({ response, isMutable: true, cacheKey, event })` which can cause cache misses because it's always writing to the mutable cache, instead of the immutable cache when protocol=`ipfs`

cache workflow needs seriously cleaned up

SAHU-01:
## Title
Fixes: #861 

## Description

The service worker was always storing responses in the mutable cache, regardless of protocol type (IPFS vs IPNS). 

- Modified `fetchAndUpdateCache()` to accept `isMutable` as a parameter
- Updated `getResponseFromCacheOrFetch()` to pass the correct `isMutable` value based on protocol:
  - `isMutable = true` for IPNS (mutable content)
  - `isMutable = false` for IPFS (immutable content)
- Now cache lookup and storage use the **same cache** for each protocol

## Notes & open questions
image

image

## Change checklist

Reviewer: @SgtPooki

### DIFF
--- a/src/sw.ts
+++ b/src/sw.ts
@@ -405,7 +405,7 @@ function getCacheKey (event: FetchEvent): string {
   return `${event.request.url}-${event.request.headers.get('Accept') ?? ''}`
 }
 
-async function fetchAndUpdateCache (event: FetchEvent, url: URL, cacheKey: string): Promise<Response> {
+async function fetchAndUpdateCache (event: FetchEvent, url: URL, cacheKey: string, isMutable: boolean): Promise<Response> {
   const response = await fetchHandler({ path: url.pathname, request: event.request, event })
   log.trace('got response from fetchHandler')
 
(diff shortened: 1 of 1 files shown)


hsanjuan:
Triage:
  * @achingbrain will look into

gammazero:
Triage: still in progres

hsanjuan:
Triage:
  * Might be fixed, needs to be re-evaluated

hsanjuan:
@SAHU-01 is this still a problem?

achingbrain:
Closing as fixed in main.  If this is still broken, please re-open with a test that fails against main.

PR #955, opened 2026-01-18
fix: reduce bundle size

achingbrain:
- Swap react/react-dom for preact
- Use Helia class from `@helia/utils` to not pull in libp2p deps
- Update various deps that have had bundle size reductions
- Enable minification

Service worker (uncompressed) 2.5MB -> 804kb
UI (uncompressed) 1.2MB -> 326kb

## Change checklist


achingbrain:
Before:

before

After:

after

achingbrain:
Initial page load before:

image

After:

image

2color:
Wow! Thanks for pushing this

PR #960, opened 2026-01-21
fix: handle service worker registration edge cases

lidel:
this PR is belt and suspenders protection against edge cases that could leave users stuck on loading screen. 

i've seen crazy bugs in Google Chrome shipping in production, and then they silently fixed them month later, wasting peoples time on debug

these should not impact normal operation, but will save debugging time if we ever hit them thanks to clear error messages and timeout

### DIFF
--- a/src/lib/register-service-worker.ts
+++ b/src/lib/register-service-worker.ts
@@ -6,19 +6,35 @@ export async function registerServiceWorker (): Promise<ServiceWorkerRegistratio
     scope: '/'
   })
 
-  // sw was registered immediately
   if (swRegistration.active?.state === 'activated') {
     return swRegistration
   }
 
+  return waitForActivation(swRegistration)
+}
+
+async function waitForActivation (swRegistration: ServiceWorkerRegistration): Promise<ServiceWorkerRegistration> {
   return new Promise((resolve, reject) => {
+    const timeoutId = setTimeout(() => {
+      reject(new Error('Service worker failed to activate within 30 seconds. Refresh the page to retry.'))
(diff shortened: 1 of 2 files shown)


achingbrain:
Is it possible to add some tests here?  I'm not sure I'd want to refactor this without having some assurance a refactor would not break something.

src/lib/register-service-worker.ts
```diff
@@ -6,19 +6,67 @@ export async function registerServiceWorker (): Promise<ServiceWorkerRegistratio
...
+ */
+async function waitForActivation (swRegistration: ServiceWorkerRegistration): Promise<ServiceWorkerRegistration> {
   return new Promise((resolve, reject) => {
-    swRegistration.addEventListener('updatefound', () => {
-      const newWorker = swRegistration.installing
-      newWorker?.addEventListener('statechange', () => {
-        if (newWorker.state === 'activated') {
-          resolve(swRegistration)
-        }
+    let settled = false
```
achingbrain: There's no need to track the settled status as a promise can either be resolved or rejected but not both.  That is, after resolving it, rejecting it has no effect and vice versa.

lidel:
@achingbrain unsure how we can test this fully, i looked at playwright docs and it cant intercept the SW registration requests in a way that allows us to test it E2E. 

Added two other tests but they don't project as much confidence as proper e2e would. 

We probably can park/close this PR and revisit if we ever see actual bugs irl. I agree we dont want to change this unless we need too.

lidel:
Closing to avoid premature optimization.

PR #967, opened 2026-01-29
chore: add issue templates with need/triage label

lidel:
ensures all new issues are assigned the need/triage label and disables blank issues to guide users through the template.


achingbrain (approved):

achingbrain:
I see there is [this too](https://github.com/ipfs/helia/pull/942) - are these the sort of files we should be distributing & keeping in sync via [github-mgmt](https://github.com/ipfs/github-mgmt/tree/master/files/.github) instead?

PR #977, opened 2026-02-17
ci: purge Cloudflare cache after deploy

lidel:
as part of HTTP caching optimization in prod/staging,  the edge and browser TTLs wil be set to 24h for faster delivery of the service worker installer on cold cache hits. this means we need to manually purge all cached assets after each deploy so edge starts serving the new version without waiting for TTL expiry. kinda best of both worlds.

the deploy pipeline now checks the Cloudflare Pages origin directly (bypasses CDN cache) before purging, then verifies the public domain serves the new version after purge.

uses separate CF tokens per environment:
- `CF_CACHE_PURGE_TOKEN_PRODUCTION` for inbrowser.link
- `CF_CACHE_PURGE_TOKEN_STAGING` for inbrowser.dev

also adds 10m timeout to the deploy job.

cc @aschmahmann @achingbrain for visibility (i'll set secrets in respective environments and then merge and test)

### Token scopes to only cache purging

> image

### Caching Rule 

For now set up in staging (`inbrowser.dev`) and not in production (`inbrowser.link`) so @aschmahmann  can compare performance –– lmk if looking good enough for me to apply to prod too.

> image

### DIFF
--- a/.github/workflows/deploy-to-production.yml
+++ b/.github/workflows/deploy-to-production.yml
@@ -18,13 +18,36 @@ jobs:
       environment: production
     secrets:
       DEPLOYMENT_GITHUB_TOKEN: ${{ secrets.UCI_GITHUB_TOKEN || github.token }}
-  wait-for-deploy:
+  # Cloudflare caching setup (configured via Cache Rules in the dashboard):
+  # - Edge TTL and Browser TTL are set to 24h for faster cold cache hits
+  # - cache key uses resolved host (not request host) and ignores query params,
+  #   because every subdomain returns the same HTML+JS service worker
+  #   installer/bootstrapper, so a single cached response serves all of them
+  #
+  # Purge all cached assets after deploy so visitors get the new version
+  # without waiting for the 24h TTL to expire.
+  purge-cache-and-verify:
+    needs: deploy
     runs-on: ubuntu-latest
+    environment: production
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v6
-      - name: Wait for deployment to complete
+      - name: Wait for Cloudflare Pages build to complete
         run: |
           # fetch the tag to ensure it's available
           git fetch origin tag ${{ inputs.tag }} --no-tags
+          # check Pages origin directly (bypasses CDN cache)
+          ./.github/scripts/wait-for-deploy.sh https://ipfs-service-worker-gateway.pages.dev ${{ inputs.tag }}
+      - name: Purge Cloudflare cache for inbrowser.link
+        run: |
+          curl --fail --request POST \
+            --url "https://api.cloudflare.com/client/v4/zones/3cf6893c26dde38efa39910a57798a59/purge_cache" \
+            -H 'Content-Type: application/json' \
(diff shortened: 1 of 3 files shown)


lidel:
I tried Snippet below, but it disables Cache Key evaluation:

Below was intended to make sure "Resolved host" from Cache Rule works the way we want, original host with content root would be forwarded in `X-Forwarded-Host`:

```js
export default {
    async fetch(request) {
      const url = new URL(request.url)
      const match = url.hostname.match(/^(.+)\.(ipfs|ipns)\.inbrowser\.dev$/)
      if (match) {
        const modifiedRequest = new Request(request, {
          headers: new Headers(request.headers),
        })
        modifiedRequest.headers.set("X-Forwarded-Host", url.hostname)
        modifiedRequest.headers.set("Host", "inbrowser.dev")
        return fetch(modifiedRequest)
      }
      return fetch(request)
    },
  }
```

Then I tried different approach with Transform Request rule, but sadly we cant use this UI to normalize Host header before Cache Key is evaluated:

> image


Cloudflare will likely not fix this, as they upsell everyone into Workers.

lidel:
Working Cloudlare Snippet that works together with Cache Rule to maximize cache hits while still allowing us to apply badbits per content root.

```js
export default {
    async fetch(request) {
      const url = new URL(request.url)
      const isSubdomain = /^.+\.(ipfs|ipns)\.inbrowser\.dev$/.test(url.hostname)

      // Skip cache key normalization for root path on subdomains: index.html
      // needs a unique cache key per subdomain so we can block and cache 410
      // for CIDs on badbits.
      //
      // For all other cases, normalize cache key to the base domain so
      // Cloudflare uses the same cache entry across all subdomains. Every
      // subdomain returns identical HTML+JS service worker installer for
      // non-root paths, so sharing cache drastically improves hit rate on
      // cold requests.
      if (isSubdomain && url.pathname === "/") {
        return fetch(request)
      }

      // Strip subdomain and query params from cache key so all subdomains
      // share the same edge cache entry for a given path.
      const cacheKey = `https://inbrowser.dev${url.pathname}`

      let response = await fetch(request, {
        cf: {
          cacheEverything: true,
          cacheKey,
          cacheTtlByStatus: {
            // Cache successful responses and client errors (including 410
            // for badbits-blocked CIDs) at the edge for 24h.
            "200-399": 86400,
            "400-499": 86400,
            // Never cache server errors -- origin might be temporarily down.
            "500-599": 0,
          },
        },
      })

      // Set browser cache TTL to match edge TTL, but skip for server errors
      // so clients retry quickly when origin recovers.
      if (response.status < 500) {
        response = new Response(response.body, response)
        response.headers.set("Cache-Control", "max-age=86400")
      }

      return response
    },
  }
```

lidel:
Merging as this PR is only about purging entire cache on deploying new version.
Snipper / Cache Rules are set at Cloudflare.

lidel:
As a smoke-test I re-deployed the same commit 00662394c286ac7cf357d14e66486c255a6786ea  to staging and cache purge worked as expected:
- https://github.com/ipfs/service-worker-gateway/actions/runs/22120739839/job/63939877656#step:4:21

PR #984, opened 2026-02-19
feat: pre-connect to routers and resolve gateway dns

achingbrain:
Use [rel=preconnect](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/preconnect) to premptively connect to configured routers, and use [rel=dns-prefetch](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/dns-prefetch) to resolve DNS for fallback trustless gateways ahead of time.

Do this in the HTML and not JS so it begins before we have installed the service worker.

- cc #979

## Change checklist

### DIFF
--- a/build.js
+++ b/build.js
@@ -3,8 +3,8 @@ import { execSync } from 'node:child_process'
 import { readFileSync } from 'node:fs'
 import fs from 'node:fs/promises'
 import path from 'node:path'
-import esbuild from 'esbuild'
 import { isIP } from '@chainsafe/is-ip'
+import esbuild from 'esbuild'
 
 /**
  * @param {string | string[]} str
@@ -182,12 +182,12 @@ const injectHtmlPages = async (metafile, revision) => {
     console.log(`Added git revision (${revision}) to ${path.relative(process.cwd(), htmlFilePath)}.`)
 
     // preconnect to routers as we will probably need to use them to find provs
-    let preconnect = new Set([
+    const preconnect = new Set([
       ...config.routers.map(toUrl)
     ])
 
     // only resolve DNS for gateways and DNS resolvers as we may use them
-    let dnsPrefetch = new Set([
+    const dnsPrefetch = new Set([
       ...config.gateways.map(toUrl),
       ...Object.values(config.dnsResolvers).map(toArray).map(toUrl)
     ])
--- a/build.js
+++ b/build.js
@@ -4,6 +4,17 @@ import { readFileSync } from 'node:fs'
 import fs from 'node:fs/promises'
 import path from 'node:path'
 import esbuild from 'esbuild'
+import { isIP } from '@chainsafe/is-ip'
+
+function toUrl (str) {
+  const url = new URL(str)
+  return `${url.protocol}//${url.host}`
+}
+
+function onlyDomains (str) {
+  const url = new URL(str)
+  return !isIP(url.hostname)
+}
 
 const copyPublicFiles = async () => {
   const srcDir = path.resolve('public')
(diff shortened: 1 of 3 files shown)


build.js
```diff
@@ -132,6 +153,21 @@ const injectHtmlPages = async (metafile, revision) => {
...
     console.log(`Added git revision (${revision}) to ${path.relative(process.cwd(), htmlFilePath)}.`)
 
+    let preconnect = [...new Set([
+      ...config.routers.map(toUrl),
+      ...Object.values(config.dnsResolvers).map(toUrl)
+    ])].map(url => `<link rel="preconnect" href="${url}" />`).join('\n    ')
+
+    let dnsPrefetch = [...new Set([
+      ...config.routers.map(toUrl),
+      ...Object.values(config.dnsResolvers).map(toUrl)
```
aschmahmann: Is this what you meant to do? From your PR description it seems like you intended this to be for gateways

build.js
```diff
@@ -132,6 +153,21 @@ const injectHtmlPages = async (metafile, revision) => {
     htmlContent = htmlContent.replace(/<%= GIT_VERSION %>/g, revision)
     console.log(`Added git revision (${revision}) to ${path.relative(process.cwd(), htmlFilePath)}.`)
 
+    let preconnect = [...new Set([
+      ...config.routers.map(toUrl),
+      ...Object.values(config.dnsResolvers).map(toUrl)
```
lidel: `dnsResolvers` could be an array so  to avoid problems in local dev maybe apply 
```suggestion
      ...Object.values(config.dnsResolvers).flat().map(toUrl)
```

here and in other places where value can be  `string | string[]` ?

build.js
```diff
@@ -132,6 +153,21 @@ const injectHtmlPages = async (metafile, revision) => {
...
+    ])].map(url => `<link rel="preconnect" href="${url}" />`).join('\n    ')
+
+    let dnsPrefetch = [...new Set([
```
lidel: i think we want to dns preload all domains from default `config` (`routers`, `dnsResolvers`, and `gateways` -- the last one is missing?)

build.js
```diff
@@ -132,6 +153,21 @@ const injectHtmlPages = async (metafile, revision) => {
     htmlContent = htmlContent.replace(/<%= GIT_VERSION %>/g, revision)
     console.log(`Added git revision (${revision}) to ${path.relative(process.cwd(), htmlFilePath)}.`)
 
+    let preconnect = [...new Set([
```
lidel: nit: we compute the same(?) list of URLs twice for preconnect and dnsprefetch, maybe do it once and reuse?

lidel (changes requested):

build.js
achingbrain: No, it was a mistake - it should be pre-connect for routers and dns prefetch for gateways and DNS resolvers.  Is fixed now.

build.js
achingbrain: It was a mistake, my thinking is to pre-connect to the routers since the provider lookup is almost inevitable (apart from identity hashes which is not common outside testing), but only dns prefetch for the gateways/dns resolvers since they may not be used.

build.js
achingbrain: If we pre-connect we shouldn't need to perform a dns prefetch since DNS resolution will occur during pre-connect, so we shouldn't need to include pre-connect domains in the dns-prefetch list.

build.js
achingbrain: Fixed!

PR #990, opened 2026-02-20
chore(ci): expose Cloudflare Pages build logs in GitHub Actions

lidel:
fetch CF Pages deployment build logs via API and write them to the job summary so contributors can see them without needing Cloudflare dashboard access.

the CF Pages API returns logs for the entire project (not per-branch), so we match deployments by commit SHA in chronological order. a shared concurrency group across staging and production deploy workflows prevents overlapping deploys from matching the wrong logs.

requires two secrets in both staging and production GitHub environments:
- `CF_ACCOUNT_ID`: Cloudflare account ID
- `CF_PAGES_READ_TOKEN`: API token with Account > Cloudflare Pages > Read

cc @achingbrain for visibility  -- im going to merge this and re-deploy staging (to the same commit)

### DIFF
--- a/.github/scripts/fetch-cf-pages-logs.sh
+++ b/.github/scripts/fetch-cf-pages-logs.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+
+# Fetches Cloudflare Pages deployment build logs via the CF API
+# and writes them to $GITHUB_STEP_SUMMARY so they are visible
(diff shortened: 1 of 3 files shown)


lidel:
Confirmed it works, we can now eyeball logs by looking at Deploy job summary at Github.

https://github.com/ipfs/service-worker-gateway/actions/runs/22231259428 → Cloudflare Pages build logs → Build log (edited)

PR #1014, opened 2026-03-24
fix: support inspecting hamt shards

achingbrain:
When a path results in a 404, retry without HAMT translation

Refs: #1066

## Change checklist


src/sw/handlers/content-request-handler.ts
```diff
@@ -265,6 +265,18 @@ async function fetchHandler ({ request, headers, renderHtml, event, logs, accept
     for (let i = 0; i < MAX_REDIRECTS; i++) {
       response = await verifiedFetch(resource, init)
 
+      // special case - if the user is trying to preview HAMT sub-shards
+      if (response.status === 404) {
```
achingbrain, line 269: This is expensive, a better solution would be for verified fetch to either somehow signal that the root CID was a HAMT shard so the `translateHAMTPath` option is available, or to add a new query string param to the service worker gateway that lets the user express their intention and set `translateHAMTPath` explicitly.

PR #1035, opened 2026-04-17
docs: improved explainer for non-browser users

lidel:
Prep for redirecting `ipfs.io` traffic here. When a non-browser client (curl, bot, RSS reader, crawler spoofing user-agent) hits the gateway, the returned HTML should be enough to self-diagnose without filing an issue or reading external docs.

Changes to `public/index.html`:

- View-source comment explains what the page is, that the browser verifies content (not the server), links to the repo, and tells users to run their own node for non-browser use cases.
- Documents the known Chromium "Save As" bug inline, with upstream link, so reports stop landing here.
- `<meta description>` rewritten for link previews (Slack, Discord, search engines).
- Added `theme-color` for mobile/PWA chrome.
- Dropped obsolete `X-UA-Compatible` (IE EOL'd 2022).

No runtime changes.


lidel:
Unsure why check fails, its unrelated to changes from this PR: 
- https://github.com/ipfs/service-worker-gateway/actions/runs/24573046022/job/71851402784?pr=1035

achingbrain:
> Unsure why check fails,

The module `@types/sinon` was accidentally specified as a production dependency by `@libp2p/tcp` - that was fixed in https://github.com/libp2p/js-libp2p/pull/3443 but it broke things here since then `@types/sinon` was not installed.

It was added as a dev dep by https://github.com/ipfs/service-worker-gateway/pull/1036 so all is good now.

achingbrain (approved):

PR #1039, opened 2026-04-29
fix: wrap renderable media in viewer page

### LINKED ISSUE #574: "Save as" broken: saves SW bootstrap HTML instead of IPFS data
> [!NOTE]
> **TLDR** summary for users finding this problem.
> This is an upstream Google Chrome bug, being tracked in https://issues.chromium.org/issues/40410035

----

Many files are not wrapped in a directory. The file extension is therefore missing.
But browsers seem to extract the file extension from the magic numbers / file signatures.

Json files
When I open a json file in ipfs.io, it gets the correct file extension.
When I open a json file in inbrowser.link its in an HTML page. I can’t download it as json.
For example:
https://ipfs.io/ipfs/bafkreieze572daxva52asutm4u2bgtjtifur2a4d2aiur7lymjcwhoeqyy
https://bafkreieze572daxva52asutm4u2bgtjtifur2a4d2aiur7lymjcwhoeqyy.ipfs.inbrowser.link/

The same applies to image files:
If I open an image in ipfs.io, the image gets the correct file extension
If I open an image file in inbrowser.link, it is in an HTML page and I cannot download it as an image from the browser.
Example:
https://ipfs.io/ipfs/bafybeiblgxritjybxffn5te6aupebq7m77xm7xiqqkiq7eete7wrblhnce
https://bafybeiblgxritjybxffn5te6aupebq7m77xm7xiqqkiq7eete7wrblhnce.ipfs.inbrowser.link/

lidel:
> Towards: 
> - https://github.com/ipshipyard/roadmaps/issues/53

> [!NOTE]
> This is an old prototype which I rebased with help of LLM after figuring out why it did not work (`<a download`). Seems to do the trick of solving UX pain + creates small surface for IPFS branding / links to docs at the top.  cc @achingbrain @aschmahmann 
> 
> Keeping as draft because it needs upstream (verifie-fetch?) fix for sniffing unixfs files that fit in a single `raw` block, see TODO in `test-e2e/media-viewer.test.ts`

## Summary

Fixes #574. Top-level navigations to renderable content (image, video, audio, PDF, text, JSON) now return a small HTML viewer with an explicit **Download** button that drives the existing `?download=true&filename=` path, so the browser saves the actual content correctly on Chromium, Firefox, and Safari.

## Why

Chromium has an open Service Worker bug ([crbug 40410035](https://issues.chromium.org/issues/40410035)) that breaks "Save As" on the gateway: right-click "Save As" on SW-served inline media saves the SW bootstrap HTML instead of the resource, and `<a download>` link clicks bypass the Service Worker entirely.

Header-only workarounds (`Content-Disposition`, `nosniff` removal, `Response` reconstruction) were all tried in #574; none help. Igalia's queued upstream fix is realistically 6 to 12 months away.

## How it works

> 

The SW wraps document-destination navigations to renderable content in an HTML viewer. The viewer embeds the same URL via ``, `<video>`, `<audio>`, or `<iframe>`; those subresource refetches arrive with `request.destination !== 'document'`, skip the wrapper, and stream raw bytes through the existing pipeline.

The Download button is `<a href="?download=true&filename=...">` *without* the HTML `download` attribute, since that attribute makes Chromium bypass the SW ([crbug 40410035](https://issues.chromium.org/issues/40410035)). The existing `?download=true` handler converts `Content-Disposition: inline` to `attachment`, and `verified-fetch` already populates `filename=` from the query param.

## Notable subtleties

- This PR rebases an old prototype I had abandoned. While picking it back up, I finally identified the cause of the original failure: `<a download>` link clicks bypass the SW on Chromium (crbug 40410035), so removing the attribute is the only thing that makes the Download button work.
- The cache key now includes `request.destination` so cached viewer HTML and cached raw bytes do not collide on the same URL.
- Unsure if preact takes case of it, but belt and suspenders escaping is necesaary, but just in case `htmlPage()` now escapes `<` in injected JSON, blocking `</script>` injection from user-controlled IPFS path segments.
- Videos don't autoplay, user needs to click play due to browser iframe/embedd limitation (otherwise all ads would play dubstep all the time, like they DID in late 90s)

## Out of scope

- HTML and SVG wrapping (would self-recurse).
- File System Access API (`showSaveFilePicker`) would improve perf. but is a progressive enhancement we can do in follow-up PR.

## Test plan

Smoke-tested locally on Chromium and Firefox with these CIDs (subdomain mode at `localhost:3000`):

- [x] image (bare CID) `bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi`
- [x] image (other) `bafybeiblgxritjybxffn5te6aupebq7m77xm7xiqqkiq7eete7wrblhnce`
- [x] image with already-extensioned path `en-wikipedia--on--ipfs-org.ipns.localhost:3000/I/page1-330px-Political_Map_of_the_Arctic.pdf.jpg.webp` 
- [x] JSON (bare CID) `bafkreieze572daxva52asutm4u2bgtjtifur2a4d2aiur7lymjcwhoeqyy`
- [ ] JSON (named path) `bafybeigxip3b5uegqlmcodhitzogr7w46uyedsai4kecrxlu4w74szyfxi/metadata.json`
- [x] text `bafybeib32tuqzs2wrc52rdt56cz73sqe3qu2deqdudssspnu4gbezmhig4`
- [x] PDF `bafybeibms6xfhs4kx3tun7dg6bjbj5obca7yj43cl5jl3pc2n7muw6ipfm`
- [x] video `bafybeig4zsnkc6caqyoj277b3fa5syvevks2542qr2pxaw7kit36yom3gy`
- [ ] same URLs with `?download=true` produce a straight download, viewer skipped
- [ ] same URLs with `?download=false` route to the existing entity inspector, no regression
- [ ] DAG-CBOR, DAG-PB, and directories route to the existing entity inspector, no regression
- [ ] Download button produces correctly-named file on Chromium, Firefox, and Safari


src/sw/lib/sniff-raw-content-type.ts
```diff
@@ -0,0 +1,77 @@
...
+ * Trustless requests (`?format=raw`, `Accept: application/vnd.ipld.raw`,
+ * CAR, dag-json, dag-cbor, etc.) are untouched: the explicit-Accept guard
+ * skips them, and other formats never carry `application/vnd.ipld.raw` to
+ * begin with.
+ *
+ * TODO: decide whether to keep this here or push the sniffing into
+ * `@helia/verified-fetch`'s `plugin-handle-raw` instead, where it would
+ * fix the spec gap for every consumer of verified-fetch. See the issue
+ * filed against helia-verified-fetch and remove this helper once the
+ * upstream fix lands.
```
lidel, line 33: @achingbrain how do you feel about it? `sniffRawContentType` could be removed if we do it upstream in
-  https://github.com/ipfs/helia-verified-fetch/pull/344

lidel: lmk if I should cleanup redundant code in this PR now that https://github.com/ipfs/helia-verified-fetch/pull/344 merged

achingbrain: All good - I've removed it all locally and it works ok, just looking to add some playwright tests that formalise the smoke tests in the issue description.

achingbrain:
Merging to deploy for experimentations, follow-up PRs may occur

PR #1049, opened 2026-05-08
fix: use ipfs path as title of dir listing

lidel:
This is  a low priority UX fix

Intention is to match Kubo's directory-listing behavior, which sets `<title>` to the content path (e.g. `/ipns/dist.ipfs.tech/kubo/`) instead of the bare CID. Only  directory  listings change; other entity previews keep the CID.

## inbrowser

```html
<title>bafybeiboqmzzqpkys2qlx5jb2rpwrq2vlab2e7epn3d5vtj77gum6suxwq</title>
```

> image

## kubo

```html
<title>/ipns/dist.ipfs.tech/kubo/</title>
```

> image


achingbrain:
Looks fine to me but it needs a test or assertions added to some existing tests.

lidel:
@achingbrain done, was unsure what is the best place, check if 99a9a8f6423ae8db802124c3dd0d3eb65349d189 and https://github.com/ipfs/service-worker-gateway/pull/1049/commits/da57a95d5763d1bcab22f7cef9506feb1b690940 make sense

PR #1063, opened 2026-05-14
docs: improve no-service-worker warning copy

lidel:
## Problem

The "Service Workers Are Not Supported" page told users what was missing (`navigator.serviceWorker is present`) but not what to do next, and only mentioned Firefox. This mentions Tor options.

## Demo

> image

cc @achingbrain  @aschmahmann for visibility

### DIFF
--- a/src/ui/pages/no-service-worker.tsx
+++ b/src/ui/pages/no-service-worker.tsx
@@ -11,16 +11,13 @@ export default function NoServiceWorkerErrorPage (): ReactElement {
           <line x1='12' y1='9' x2='12' y2='13' />
           <line x1='12' y1='17' x2='12.01' y2='17' />
         </svg>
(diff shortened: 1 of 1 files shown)


achingbrain:
> body bumped to f4 for legibility.

I am not averse to this change but now this page is inconsistent with the others.  We should change this everywhere or nowhere.

lidel:
Triage: trim down, frame towards Tor Browser user (as its realistically the only one without SW API)

lidel:
@achingbrain rebased on top of #1115, kept font sizes from main, and [simplified message](https://github.com/user-attachments/assets/fa65883e-3a84-49d6-887f-7b2381b02b2c), mind taking another look?

achingbrain (approved):

PR #1064, opened 2026-05-14
feat: enable webtransport on all supporting browsers

lidel:
## What was broken
A Firefox useragent sniff in `src/sw/lib/verified-fetch.ts` kept Chrome, Edge, and Safari users from dialing WebTransport peers, and stripped `webtransport` from the `filter-addrs` query sent to `delegated-ipfs.dev`.

## Fix
Feature-detect `globalThis.WebTransport`. WebTransport reached Baseline with Safari 26.4 (March 2026), so Chrome 97+, Edge 98+, Firefox 114+, and Safari 26.4+ now dial WebTransport peers and the routing query includes `webtransport`.

## Caveat: Chromium reliability
Chromium WebTransport throttles after enough sessions accumulate, blocking new dials until the page reloads (upstream: [chromium#40069954](https://issues.chromium.org/issues/40069954); IPFS impact: ipfs/in-web-browsers#211). 

We believe that shipping this despite that being wip is still net-positive: peers that are only reachable over WebTransport become dialable in Chrome/Edge, and libp2p falls back to other transports (WebSockets, HTTPS) or peers when WebTransport fails.

cc @achingbrain

### DIFF
--- a/src/sw/lib/verified-fetch.ts
+++ b/src/sw/lib/verified-fetch.ts
@@ -39,8 +39,9 @@ async function libp2pDefaults (): Promise<Libp2pOptions> {
     webSockets()
   ]
 
-  // WebTransport only works reliably in Firefox at the time of writing
-  if (navigator?.userAgent?.toLowerCase().includes('firefox') === true) {
+  // Enable WebTransport when the browser exposes the API.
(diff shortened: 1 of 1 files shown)


achingbrain:
Really `@libp2p/webtransport` should fail gracefully if there's no implementation present but this will do until then.  Thanks!

PR #1081, opened 2026-05-18
fix(snippets): installer cache poisoning

lidel:
> [!NOTE]
> This needs to be smoke-tested on `inbrowser.dev`, but should be safer than the current status quo 

A transient 3xx from origin gets pinned for 24h under the per-subdomain installer cache key, putting every URL on the subdomain into a redirect loop at one PoP. Seen today ([thread](https://ipshipyard.slack.com/archives/C06CY2RQ50V/p1779107069973919?thread_ts=1779095175.181589&cid=C06CY2RQ50V)) after #1061: rainbow returned `308 Location: /` for `bafy...ipfs.inbrowser.link` and the BOS PoP cached it (`cf-cache-status: HIT`, age 5640s).

Make the cache key honest:

* installer branch: GET and HEAD fetch `/` with `Accept: text/html` and `redirect: manual`, ignoring the requested path. `cacheTtlByStatus` restricted to 2xx (`200-299`).
* `/ipfs-sw-*` branch: `redirect: manual` added so a followed 3xx cannot land 2xx content under the asset cacheKey.

Zone purge on `inbrowser.link` after deploy to flush existing poisoned entries (~22h remaining on TTL otherwise).

### DIFF
--- a/src/cloudflare/snippets/02_shared_sw_installer_cache.ts
+++ b/src/cloudflare/snippets/02_shared_sw_installer_cache.ts
@@ -70,6 +70,8 @@ export default {
           cacheKey,
           cacheTtlByStatus: {
             '200-299': EDGE_CACHE_TTL_S,
+            // Do not cache errors. A transient failure must not be stuck
+            // in cache for the full TTL.
             '300-599': 0
           }
         }
@@ -103,6 +105,8 @@ export default {
         cacheKey: `https://${url.hostname}/__sw_installer_html`,
         cacheTtlByStatus: {
           '200-299': EDGE_CACHE_TTL_S,
+          // Do not cache errors. A transient failure must not be stuck
+          // in cache for the full TTL.
           '300-599': 0
         }
       }
--- a/src/cloudflare/snippets/02_shared_sw_installer_cache.ts
+++ b/src/cloudflare/snippets/02_shared_sw_installer_cache.ts
@@ -56,7 +56,15 @@ export default {
       const baseDomain = url.hostname.split('.').slice(-2).join('.')
       const cacheKey = `https://${baseDomain}${url.pathname}`
 
-      return fetch(request, {
+      // redirect: manual matches the installer branch. With the
+      // default redirect: follow, an origin 3xx on an asset path
+      // would be followed and the target's 2xx body would land in
+      // cache under the asset cacheKey, serving wrong content.
+      // cacheTtlByStatus 300-599: 0 only drops the 3xx itself,
+      // not the followed 2xx.
+      const assetReq = new Request(request, { redirect: 'manual' })
+
+      return fetch(assetReq, {
         cf: {
           cacheEverything: true,
           cacheKey,
--- a/src/cloudflare/snippets/02_shared_sw_installer_cache.ts
+++ b/src/cloudflare/snippets/02_shared_sw_installer_cache.ts
@@ -1,22 +1,50 @@
 // Cloudflare Snippet: shared service worker installer cache
 //
-// Controls edge caching for the service worker gateway. Every subdomain
-// serves the same installer HTML and the same versioned JS/CSS.
+// What this snippet does
 //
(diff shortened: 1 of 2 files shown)


test/cloudflare/snippets/02_shared_sw_installer_cache.spec.ts
```diff
@@ -1,200 +1,181 @@
...
+      const { request } = await callHandler('https://bafyxxx.ipfs.inbrowser.dev:8443/anything')
+      expect(request!.url).to.equal('https://bafyxxx.ipfs.inbrowser.dev:8443/')
     })
 
-    it('HTML and versioned SW asset TTLs match (avoid HTML-JS skew)', async () => {
-      const { cf: cfHtml } = await callHandler('https://bafyxxx.ipfs.inbrowser.dev/page')
-      const { cf: cfAsset } = await callHandler('https://bafyxxx.ipfs.inbrowser.dev/ipfs-sw-main.js')
-      expect(cfHtml.cacheTtlByStatus['200-399']).to.equal(cfAsset.cacheTtlByStatus['200-299'])
+    it('rewrites HEAD to HEAD "/" and shares the GET cache key', async () => {
+      const { request: gReq, cf: gCf } = await callHandler('https://bafyxxx.ipfs.inbrowser.dev/page')
```
achingbrain: `gReq` is not used - did you mean to assert on the value?

lidel: no, scar tissue, removed in d02041d8aea5753c4c609f14277219ad0daeb7bb

src/cloudflare/snippets/02_shared_sw_installer_cache.ts
```diff
@@ -28,31 +56,54 @@ export default {
...
+      const assetReq = new Request(request, { redirect: 'manual' })
+
+      return fetch(assetReq, {
         cf: {
           cacheEverything: true,
           cacheKey,
           cacheTtlByStatus: {
             '200-299': EDGE_CACHE_TTL_S,
-            // Do not cache errors. A transient failure must not be stuck
-            // in cache for the full TTL.
```
achingbrain, line 38: Why remove the comment?  It gives useful context.

src/cloudflare/snippets/02_shared_sw_installer_cache.ts
```diff
@@ -28,31 +56,54 @@ export default {
...
-    return fetch(request, {
+    return fetch(installerReq, {
       cf: {
         cacheEverything: true,
-        cacheKey,
+        cacheKey: `https://${url.hostname}/__sw_installer_html`,
         cacheTtlByStatus: {
-          '200-399': EDGE_CACHE_TTL_S,
-          // Do not cache errors. A transient failure must not be stuck
-          // in cache for the full TTL.
```
achingbrain, line 54: Why remove the comment? It gives useful context.

test/cloudflare/snippets/02_shared_sw_installer_cache.spec.ts
```diff
@@ -1,200 +1,181 @@
-// Tests for 02_shared_sw_installer_cache.js Cloudflare Snippet
```
achingbrain, line 1: There's a lot of noise in these tests.  The additions are good, but there are renames and deletions - restoring some of the deleted tests shows they would still pass so should probably have been left alone?

Is this an LLM or was it intentional?

achingbrain (changes requested):

test/cloudflare/snippets/02_shared_sw_installer_cache.spec.ts
lidel: Not intentional, I was in a rush this morning and generated tests with wrong preset. 

Fixed in c1a8e071edc4fad6dea6d26b2eb0764ee76c8b95 the smaller diff should be easier to review.

achingbrain (approved):

PR #1089, opened 2026-05-21
fix: make url hash available to loaded HTML page

### LINKED ISSUE #1088: HashRouter consumes hash routes meant for the hosted app
> Reported on the [IPFS forum](https://discuss.ipfs.tech/t/dweb-link-now-redirects-to-inbrowser-link/20158/3); observed on inbrowser.link running v3.4.3 (`HEAD@9ea703c`). cc @achingbrain

The gateway UI uses react-router's `HashRouter` (`src/ui/index.tsx`). When a hosted subdomain app does its own hash-based routing, the gateway's first load (before the service worker takes over) tries to match the app's hash and logs `No routes matched location '/mail/Inbox'`.

**Steps to reproduce:**

1. Host an SPA that uses URL-hash routing.
2. Open it at a deep route on a subdomain gateway, e.g. `https://<cid>.ipfs.inbrowser.link/#/mail/Inbox`.
3. The console shows `No routes matched location '/mail/Inbox'` from the gateway bootstrap.

**Expected:** the gateway leaves hash fragments alone when they belong to the hosted app, so SPAs that route on the hash keep working. 

Can we make it so only specific hash fragments specific to ipfs-sw- are captured, and everything else is ignored?

achingbrain:
Fixes two bugs:

1. Uses `window.location.reload` when handing off to the service worker instead of assigning `window.location` - if a hash is present in the url the former works, the latter doesn't
2. Make the page index route also a wildcard so appropriate UI is shown while loading the content

Fixes #1088

## Change checklist

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -164,8 +164,15 @@ async function main (): Promise<void> {
         await registerServiceWorker()
       }
 
+      // remove service worker nav to prevent endless redirects
+      if (window.location.hash.startsWith('#/ipfs-sw')) {
+        window.location.hash = ''
+      }
+
       // reload so the subdomain request is handled by the service worker
-      window.location.href = url.toString()
+      // (including the hash)
+      window.location.reload()
+
       showUIAfterDelay(request)
       return
     }
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -164,14 +164,56 @@ async function main (): Promise<void> {
         await registerServiceWorker()
       }
 
-      // remove service worker nav to prevent endless redirects
+      // Clear SW-UI hashes before navigating: otherwise the post-reload
+      // React app would match a `/ipfs-sw-…` route from the HashRouter
+      // and show a SW UI page instead of the requested CID content,
+      // causing endless redirects between the two.
       if (window.location.hash.startsWith('#/ipfs-sw')) {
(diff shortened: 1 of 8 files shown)


lidel:
Triage note: 
- test with PDF and `#page=1` and see if media viewer (save-as workaround) still works

lidel (approved):
Pushed  68b3042  on top.

- SW strips `url.hash` at the fetch boundary: Chromium leaks the nav hash via `event.request.url` on `reload()`, polluting cache keys and producing `URL#frag#frag` iframe srcs.
- Iframe forwards `globalThis.location.hash` so PDFs honor `#page=N` and text anchors work.
- Bootstrap reload branches on hash: no-frag stays on `href = url.toString()` so the browser revalidates from cache (this way we avoid extra Cloudflare hits in the common path, and we dont drain the quota we have too fast), frag falls back to `reload()`.


@achingbrain  If it looks fine, feel free to ship.


## Demo

http://bafybeibms6xfhs4kx3tun7dg6bjbj5obca7yj43cl5jl3pc2n7muw6ipfm.ipfs.localhost:3000/#page=6

> image

achingbrain:
Added an extra test for vanilla websites as reported in #1088 to guard against future regressions.

PR #1109, opened 2026-06-03
fix: respect cache-control response header

achingbrain:
Removes the custom 1h `sw-cache-expires` header added to responses in https://github.com/ipfs/service-worker-gateway/pull/92 instead deriving the cached response staleness/freshness based on the received `Date` (defaulting to `now` if not set), any `Age` header, and the `Cache-Control` or `Expires` headers according to RFC 9111.

Resorts to the previous default of 1h if it's not possible to detect the server-supplied max-age.

## Change checklist

### DIFF
--- a/src/sw/handlers/content-request-handler.ts
+++ b/src/sw/handlers/content-request-handler.ts
@@ -7,6 +7,7 @@ import { errorToObject } from '../../lib/error-to-object.ts'
 import { getSubdomainParts } from '../../lib/get-subdomain-parts.ts'
 import { getSwLogger } from '../../lib/logger.ts'
 import { isBitswapProvider, isFallbackTrustlessGatewayProvider, isTrustlessGatewayProvider } from '../../lib/providers.ts'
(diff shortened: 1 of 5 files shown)


lidel (approved):
LGTM, reading the TTL from the response `cache-control` headers is the right call, and it works well with verified-fetch already turning the IPNS record TTL and DNSLink DNS TTL into `max-age`. 

iiuc the worker now caches mutable content for its real TTL instead of a fixed one hour. That's where ipfs/boxo#329 has been heading (and we still need to fix the GO side, but cool that JS already does the right thing herE).

FYSA I pushed a few small nits while reading through the caching code. Nothing blocking, just nits / spec-compliance, the details are in the commit messages:

- 70d41cf Expires TTL was in milliseconds but read as seconds.
- 02a4b59 the `Age` header got counted twice.
- 437fd0b guard a `cache-control` with no `max-age` so it doesn't throw.
- 74255fc lowercase directive names so casing variants still match.
- 626d298 let fresh `must-revalidate` responses stay fresh.
- 7ed1c7a / 7fcd125 drop a dead log line and reuse the directive parser for `no-store`.

PR #1151, opened 2026-07-07
fix: remove cache on failed asset get

achingbrain:
When the service worker is older than the currently deployed version, it will generate and have cached HTML pages that contain script tags with src attributes that point to old assets that have been removed from the server (e.g. after a code change `/ipfs-sw-index-HASH-A.js` becomes `/ipfs-sw-index-HASH-B.js`.

Detect when this happens, remove old caches, unregister the service worker, and reload the page so we get the updated versions.

## Change checklist


achingbrain:
Merging to test fix in dev

PR #1156, opened 2026-07-10
fix: cyclic reload from a poisoned cdn cache

### LINKED ISSUE #1155: Cyclic reload
### Checklist

### Description

https://github.com/user-attachments/assets/b8a35e75-7e12-4438-8cce-92dea109009e

lidel:
## Problem

iiuc after the 3.4.10 deploy, every page on inbrowser.link was a blank screen (#1155).
it got fixed via Purge Everything in Cache at Cloudflare, but may come back.

A browser asked for a hashed chunk the origin did not have yet, and the `/* /index.html 200` catch-all answered with installer HTML and a 200 rather than a 404. A 200 is cacheable, so the edge pinned that HTML for 24h under a cache key shared by every subdomain. Browsers reject a module on its MIME type, the bootstrap read that as a new version shipping and reloaded, about 100 times in 12 seconds, until Cloudflare rate limited us. No error, just a spinner.

## Fix

The origin stops lying, the snippet stops passing a lie on, and the bootstrap stops reloading into one. Worst case is now a single wasted reload and a page that tells you to wait, rather than a loop that costs wasted requests.

01-desktop-light

## Changes

@achingbrain since the problem is nuanced, and fixes are at different abstraction layers, it may be easier to review in stages. Each commit stands alone, in this order:

- 97caad9 **fix: 404 missing versioned assets**
The root cause. Drops the catch-all so Pages, `main.go` and the dev server all answer a missing `/ipfs-sw-*` with a real 404, which the snippet's `cacheTtlByStatus: {'300-599': 0}` already refuses to cache. Nothing poisonable gets written. The catch-all served nobody: the snippet answers non-asset paths by fetching `/` anyway.

- 9052ba5 **fix: never serve html under an asset url**
Insurance for the cutover, when the deployment being replaced still carries the old catch-all. The snippet checks the content type; a `text/html` body under a `.js` URL means the file is gone, so it refetches origin with `cache: 'no-store'` and serves the real chunk, or answers 404. A poisoned entry costs one fetch, not a day.

- 84987ff **fix: retry a failed app chunk once**
Caps the loop. One purge and reload clears every cache the client owns; if the chunk is still missing, nothing a reload can reach will fix it, so the second failure renders the page above. The marker lives in `history.state`, so it survives the reload and still works when the browser blocks storage.

- e41507c **ci: purge cache even if deploy is cut short**
In the bad deploy the purge did run, and the cache refilled with the poison five seconds later while Pages was mid rollout. The job then timed out and the run went down as cancelled. Now the purge runs even when the job is cut short, and again a minute later once Pages has propagated.

We need to test this end-to-end on `.dev` as I was not able to test fully.

Note: a subdomain can still hold stale installer HTML for 24h. A double unconditional total global Cloudflare cache purge clears that, and the 404 keeps it from becoming HTML served as a script.

---

- Closes #1155

### DIFF
--- a/.github/scripts/purge-cache.sh
+++ b/.github/scripts/purge-cache.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+# Purge the whole Cloudflare cache for a zone.
+#
+# purge_everything is the only purge available below Enterprise: purging by
+# prefix or by cache tag is not, and single-file purge by URL does not reach
+# entries stored under the custom cache keys that Snippet
+# 02_shared_sw_installer_cache sets. So this is a blunt instrument by
+# necessity.
+#
+# It is safe to run more than once, and safe to run when a deploy has not
+# landed. Built assets are content hashed, so the worst a purge can do is leave
+# the cache cold for a moment.
+
+set -euo pipefail
+
+: "${CF_ZONE_ID:?CF_ZONE_ID must be set}"
+: "${CF_TOKEN:?CF_TOKEN must be set}"
+
+echo "Purging Cloudflare cache for zone ${CF_ZONE_ID}"
+
+curl --fail --silent --show-error --request POST \
+  --url "https://api.cloudflare.com/client/v4/zones/${CF_ZONE_ID}/purge_cache" \
+  -H 'Content-Type: application/json' \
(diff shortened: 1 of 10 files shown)


achingbrain:
Seems fair enough. It's very hard to test locally so I'll merge & deploy to dev.

PR #1161, opened 2026-07-14
fix: wait for UI before reloading in Safari

### LINKED ISSUE #1159: Safari sometimes triggers 310 on service worker install
### Checklist

### Description

On first load of a subdomain gateway page, if no service worker is present on the domain we install the service worker, wait for it to activate then reload the page so it can be handled by the service worker and the content resolved from the IPFS network.

There seems to be a bug in Safari whereby the service worker doesn't report itself as installed or active after the reload, meaning we go back to the beginning and install the service worker, wait for it to be active and reload.

This can happen four or five times before the service worker reports it's status successfully.

Seems to happen on desktop and mobile Safari currently against both inbrowser.link and inbrowser.dev so is unrelated to the recent unreleased changes around detecting out of date service workers.

achingbrain:
In Safari loading a script during navigation triggers the onerror event listener which unregisters the service worker and reloads the page.

Instead wait for the "Your download is happening" page to appear before reloading in Safari only.

Fixes #1159

## Change checklist

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -34,14 +34,14 @@ async function main (): Promise<void> {
   if (!isBrowserSupported()) {
     // browser is missing Web APIs we need, render the UI which will tell the
     // user their browser is unsupported
-    await renderUi()
+    renderUi()
     return
   }
 
   if (!('serviceWorker' in navigator)) {
     // no service worker support, render the UI which will tell the user
     // service workers are not supported
-    await renderUi()
+    renderUi()
     return
   }
 
@@ -60,7 +60,7 @@ async function main (): Promise<void> {
     globalThis.originIsolationWarning = {
       location: window.location.href
     }
-    await renderUi()
+    renderUi()
     return
   }
 
@@ -80,7 +80,7 @@ async function main (): Promise<void> {
       logs: []
     }
 
-    await renderUi()
+    renderUi()
     return
   }
 
@@ -106,15 +106,16 @@ async function main (): Promise<void> {
       logs: []
     }
 
-    await renderUi()
+    renderUi()
     return
   }
 
   try {
     if (request.type === 'path' || request.type === 'native') {
+      await showDownloadingPageAfterDelay(request)
+
       // redirect to subdomain
       window.location.href = request.subdomainURL.href
-      showUIAfterDelay(request)
       return
     }
 
@@ -137,8 +138,8 @@ async function main (): Promise<void> {
           const request = parseRequest(uri, new URL(globalThis.location.href))
 
           if (request.type === 'subdomain' || request.type === 'path' || request.type === 'native') {
+            await showDownloadingPageAfterDelay(request)
             window.location.href = request.subdomainURL.href
-            showUIAfterDelay(request)
             return
           }
           return
@@ -182,6 +183,8 @@ async function main (): Promise<void> {
         url.pathname = ''
       }
 
+      await showDownloadingPageAfterDelay(request)
+
       // Trigger a navigation so the just-installed service worker can
       // intercept the subdomain request. Pick the cheapest primitive
       // that still works in the presence of a URL fragment.
@@ -224,8 +227,6 @@ async function main (): Promise<void> {
         window.location.reload()
       }
 
-      showUIAfterDelay(request)
-
       return
     }
   } catch (err: any) {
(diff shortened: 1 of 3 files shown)


achingbrain:
Merging to deploy to dev.

src/index.tsx
```diff
@@ -325,16 +327,20 @@ function showAppChunkError (): void {
...
     const script = document.createElement('script')
-
     script.addEventListener('load', () => {
       forgetChunkRetry()
     })
-
     script.addEventListener('error', () => {
+      // Safari errors when loading a script during page navigation so swallow
+      // the error if it fails
+      if (isWebkit()) {
```
lidel, line 340: hm.. this gates on the engine rather than on whether a navigation is in flight, so it also swallows real chunk failures on paths that never navigate: the landing page, the 310 page, the final `renderUi()` fallthrough. #1155 -style poisoning hits every engine, and on iOS every browser is WebKit, so those users lose both the one-shot recovery and the error page while Chrome/Firefox keep them.

as a precaution, since `main()` starts every navigation itself, a flag can scope this to the actual failure mode:

```ts
let navigating = false
// set to true right before each `window.location.href = ...` / `location.reload()`,
// then here, instead of (or in addition to) the engine check:
if (navigating && isWebkit()) {
  return
}
```

With the pre-navigation render the chunk loads while `navigating` is still false, so real failures keep recovering (which also unblocks the hang flagged at `waitForUiToRender()`), and Safari's spurious error fires after the assignment, so it stays ignored. A `pageshow` listener resetting the flag covers bfcache restores.

src/index.tsx
```diff
@@ -432,27 +438,72 @@ function tooManyRedirects (storageKey: string, maxRedirects = 5, period = 5_000)
...
+ * Detect WebKit browsers
+ */
+function isWebkit (): boolean {
+  return 'GestureEvent' in globalThis
+}
+
+/**
+ * Wait for the UI to be present in the DOM
+ */
+async function waitForUiToRender (): Promise<void> {
```
lidel, line 494: This wait has no timeout, and with the WebKit early-return in the error listener nothing can end it: if the chunk 404s (the #1155 case), there is no retry, no error page, no `<header>`, and the navigation below the `await` never runs. Same hang via CSS: `injectCSS()` only listens for `load`, so a blocked stylesheet stalls the UI before `root.render`.

Maybe add a deadline here (navigate anyway after a few seconds) and an `error` listener in `injectCSS()` that resolves?

src/index.tsx
```diff
@@ -432,27 +438,72 @@ function tooManyRedirects (storageKey: string, maxRedirects = 5, period = 5_000)
...
- * on a loading page
+ * on a loading page.
+ *
+ * If the user is on a WebKit-based browser loading the UI js file while a
+ * navigation is occurring will fail so wait for the UI to appear before
+ * redirecting, otherwise load the UI asynchronously.
  */
-function showUIAfterDelay (request: ResolvableURI): void {
+async function showDownloadingPageAfterDelay (request: ResolvableURI, delay = 500): Promise<void> {
+  if (isWebkit()) {
```
lidel, line 448: nit: on WebKit this now loads the full UI (~100 kB gzipped plus CSS) and shows the downloading page before every content navigation, twice for path gateway URLs (root origin, then subdomain). 

A minimal notice built from bootstrap DOM, the way `showAppChunkError()` does it, would render instantly and not depend on the chunk loading at all.

src/sw/pages/page.ts
```diff
@@ -45,30 +45,124 @@ globalThis.${key} = ${JSON.stringify(value, null, 2).replace(/</g, '\\u003c')}
...
+      // needs no storage permission, so the cap still holds in a browser
+      // configured to deny block storage
+      }
+      const CHUNK_RETRIED = 'ipfsSwChunkRetried'
+
+      function alreadyRetriedChunk () {
+        return history.state?.[CHUNK_RETRIED] === true
+      }
+
+      function markChunkRetried () {
```
lidel, line 63: unsure how real in practice, but this cap can get lost: the recovery unregisters the SW, the bootstrap then re-navigates fragment-less URLs with `location.href = url.toString()`, and Firefox drops `history.state` on that navigation (Chromium keeps it, WebKit untested; checked with a Playwright script). After that the purge/unregister/reload loop runs uncapped again.

Fix idea: also write the marker to `sessionStorage` and check both places. `sessionStorage` survives the navigation, and the `history.state` copy still covers browsers that block storage:

```js
function markChunkRetried () {
  history.replaceState({ ...history.state, [CHUNK_RETRIED]: true }, '')
  try { sessionStorage.setItem(CHUNK_RETRIED, '1') } catch {}
}

function alreadyRetriedChunk () {
  try { if (sessionStorage.getItem(CHUNK_RETRIED) != null) return true } catch {}
  return history.state?.[CHUNK_RETRIED] === true
}
```

(`forgetChunkRetry()` clears both.)

src/sw/pages/page.ts
```diff
@@ -45,30 +45,124 @@ globalThis.${key} = ${JSON.stringify(value, null, 2).replace(/</g, '\\u003c')}
...
+        document.querySelector('.loading-indicator-js')?.classList.add('hidden')
+
+        const root = document.getElementById('root')
+
+        if (root == null) {
+          return
+        }
+
+        ${''
+        // .sw-error is defined in the inline stylesheet of index.html, because
```
lidel, line 90: iiuc this page ships no stylesheet (it's all inside the chunk that just failed), so `.sw-error` matches nothing and the message renders unstyled; the `.loading-indicator-js` lookup is a no-op too. Maybe inline the few rules here the way #1156 did in the bootstrap page (https://github.com/ipfs/service-worker-gateway/blob/9e0c642ee6a69ed603a941970378b28eec849140/public/index.html#L121-L152) and/or update this comment?

lidel (commented):
Sorry for the slow review @achingbrain, this landed while I was writing it up but got distracted and never submitted 🙈 . Leaving the comments in case they help; the two on `src/index.tsx` (a possible hang, and #1155 recovery lost on WebKit) are the ones worth a look, and if they don't reproduce in real Safari on inbrowser.dev, feel free to disregard.

Not covered inline: CI never runs the safari project (`main.yml` calls `test:webkit`, but the script is `test:safari`). Playwright polyfills `GestureEvent` (microsoft/playwright#22768), so once the name matches those branches actually run; an e2e that 404s the app chunk would cover both index.tsx issues.

PR #1193, opened 2026-08-20
chore: use .net gw for fallback

lidel:
Switches the recursive gateway fallback from `.link` to `.net`, which is provisioned for this project alone rather than as a general-purpose public gateway.

cc @achingbrain @aschmahmann @cewood


.github/workflows/dnslink-smoke-tests.yml
lidel, line 1: ℹ️  Cleanup of dead workflow. This has been failing since we switched to Cloudflare Pages, removing rather than fixing TLDs in this workflow.

achingbrain: Yeah, I've been meaning to remove this 👍

